PR #233, opened 2020-07-03
[CHEC-667] - Bugfix/purge unused styles

### LINKED ISSUE #145: Tailwind is not purging unused styles because no template paths have been provided
image

Happens when running `yarn build-lib`

https://tailwindcss.com/docs/controlling-file-size/#removing-unused-css

jaepass:
- Added purge option to tailwind.config and included all template paths using custom tailwind classes

Closes #145


tailwind.config.js
```diff
@@ -16,6 +16,17 @@ const fontSizes = {
...
 };
 module.exports = {
+  purge: {
+    enabled: process.env.NODE_ENV === 'production',
+    content: [
+      './components/**/*.vue',
+      './components/*.vue',
+      './components/*.js',
+      './stories/**/*.mdx',
+      './stories/**/**/*.mdx',
```
ScopeyNZ, line 26: I don't think we should include these - as tailwind classes used here shouldn't be part of our bundle.

ScopeyNZ:
Hmmmn. Unfortunately that breaks the deploy preview storybook. So maybe we need to set the purge paths differently when building the storybook? This has gotten a little challenging.

jaepass:
hmmmmm is right, will continue to investigate!

ScopeyNZ:
Okay I'm not sure how we can do this. We don't want to include stories in our configuration when building the bundle (`yarn build-lib`) but we do want to include it when building the storybook (`yarn storybook:build`). Maybe there's a way you can determine what command is running with an environment variable?

tailwind.config.js
```diff
@@ -16,6 +16,16 @@ const fontSizes = {
   '5xl': '2.5rem',
 };
 module.exports = {
+  purge: {
+    enabled: process.env.NODE_ENV === 'production',
```
ScopeyNZ, line 20: I _think_ Vue CLI takes care of enabling this for you in production already, so I don't think you need this line.

ScopeyNZ: (Might be wrong)

jaepass: yes you're right i did come across this, it does by default. after some digging, think the issue is that the tailwind extractor does not include mdx ext.

will try your above suggestion with seeing which command is running

ScopeyNZ:
Just disabling the purgecss for storybook building is probably an acceptable workaround too.

jaepass:
wait so looks like PurgeCss was added as a built-in in v1.4 which should then allow for the purge option in the configuration. https://github.com/tailwindcss/tailwindcss/releases
https://github.com/tailwindcss/tailwindcss/pull/1639
let me know if i should go ahead and upgrade to 1.4

ScopeyNZ:
Oh right. That's probably worth doing. We've already updated dashboard a while back.

jaepass:
oh shucks its still purging storybook...

tailwind.config.js
```diff
@@ -15,7 +15,17 @@ const fontSizes = {
   '4xl': '2rem',
   '5xl': '2.5rem',
 };
+
 module.exports = {
+  purge: {
+    enabled: process.env.PURGE_ENABLED,
```
ScopeyNZ, line 21:
```suggestion
    enabled: Boolean(process.env.PURGE_ENABLED),
```

ScopeyNZ:
My bad - that should probably fix it.

jaepass:
all good now thanks for you help on this!

jaepass:
you're going to laugh to know i spent a good portion of last night adding those whitelist elements... lol

PR #237, opened 2020-07-09
[CHEC-736] Feature/Add new icons

jaepass:
- added marketing icons that haven't been added
- added new docs icons


src/lib/icons.js
```diff
@@ -94,4 +116,26 @@ export const mIcons = {
...
   enterprisecircle: IcLrgEnterprise,
+  start: IcColorStart,
+  cart: IcColorCart,
+  checkout: IcColorCheckout,
+  orders: IcColorOrders,
+  products: IcColorProducts,
+  categories: IcColorCategories,
+  backend: IcLrgBackend,
+  business: IcLrgBusiness,
+  'color-checkout': IcLrgCheckout,
```
robbieaverill, line 127: Should this be using `IcColorCheckout` instead of `IcLrgCheckout`?

jaepass: ah yea thats confusing, it's supposed to use IcLrgCheckout, will change label to 'checkout-lrg'!

ScopeyNZ: `lg` or `large` is preferable. `lg` because that's what we use with all the tailwind stuff, or `large` because it's the actual word.

jaepass:
edited label to 'checkout-lg' and added social icons as @john-raymon is working on the docs footer

ScopeyNZ:
Yep it's `lg`.

jaepass:
oh shoot was renaming all prefixes, but no big deal now can do later on! thanks!

PR #243, opened 2020-07-16
[CHEC-796] Updating Modal Overlay 

kvisca:
- Added Box Shadow
- Added light overlay with option to use either light or dark.

Note: 
This PR does not include removing the close button as I need to verify with the design first. Since we are removing the ability to close by clicking off the modal we may be getting ourselves into a bit of a UX footgun.

### DIFF
--- a/src/components/ChecModal.vue
+++ b/src/components/ChecModal.vue
@@ -89,7 +89,8 @@ export default {
   &__overlay {
     @apply fixed top-0 left-0 w-full h-full z-50 flex justify-center items-center overflow-scroll;
     background-color: rgba(255, 255, 255, 0.95);
-    &--dark{
+
+    &--dark {
       background-color: rgba(65, 85, 108, 0.9);
     }
   }
--- a/src/components/ChecModal.vue
+++ b/src/components/ChecModal.vue
@@ -1,7 +1,6 @@
 <template>
   <component
     :is="form ? 'form' : 'div'"
-    class="modal__overlay"
     :class="classObject"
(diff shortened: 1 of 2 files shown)


kvisca:
@blazepollard and I discussed the close buttons and have decided that they will be staying.

src/components/ChecModal.vue
```diff
@@ -1,5 +1,9 @@
 <template>
-  <component :is="form ? 'form' : 'div'" class="modal__overlay" @click.self="emitClose">
+  <component
+    :is="form ? 'form' : 'div'"
+    :class="classObject"
```
ScopeyNZ, line 4: Could've just made this:

```
    :class="`modal__overlay modal__overlay--${this.overlay}`"
```

but 🤷‍♂️ .

src/components/ChecModal.vue
```diff
@@ -73,12 +88,15 @@ export default {
 .modal {
   &__overlay {
     @apply fixed top-0 left-0 w-full h-full z-50 flex justify-center items-center overflow-scroll;
-    // Special / overlay
```
robbieaverill, line 76: We had this comment here deliberately, because otherwise we'd have expected to use a colour from a variable or Tailwind mixiny thing - could you put the comment back, and apply it to the new dark colour you added on line 94 please?

src/components/ChecModal.vue
kvisca: Updated based on feedback

Issue #231, opened 2020-07-02
Clicking off drop downs in modals can lead to accidental closes


kvisca:
**Issue:**
If a user opens a dropdown in a modal the typical user will click out to close that dropdown. If the user mistakenly clicks outside the modal bounds, the user will close the modal and all work will be lost. 

**Solution:**
Catch the click event so the dropdown closes instead of the modal.

ScopeyNZ:
Talking with @dvnkshl - we will just disable clicking the overlay to dismiss.

ScopeyNZ:
This will also involve going through and checking all existing modals already have an option that's not a the provided close button or clicking the overlay.

ScopeyNZ:
We've merged #244 which removes the overlay click. The comment I made around going through and checking modals has been left out as removing the close button on each modal is still up in the air. I'll add that A/C to any task that removes the close button in the future (if any)

PR #244, opened 2020-07-16
[CHEC-795] Removing ability to close on overlay click

kvisca:
- Removing the ability to close on overlay click
- Removing story that demonstrated this. 

Closes issue #231


robbieaverill:
Hey team, don't forget to update the dashboard when making backwards incompatible changes like this. We have an automated test for this behaviour in the dashboard which is now failing =)

PR #245, opened 2020-07-16
[CHEC-790] Updating card header variant to new style

kvisca:
Updating the card variant of the header to match the new designs.

Screen Shot 2020-07-16 at 4 02 42 PM


ScopeyNZ:
Labels in the header (as shown in your example) feel a bit oversized now - but that's a problem for @blazepollard to consider. Let's go with this for now and change labels if they come up.

PR #246, opened 2020-07-21
[CHEC-797] Fixing accordion to not clip card shadow.

kvisca:
**Issue:**
Accordions rely on `overflow: hidden` on the body to work, this is causing the cards box-shadows to be clipped when placed inside of an accordion.

**Changed:**
- Added `index.stories.js` to the `.eslintingnore` due to an issue that required you to save the file each time you saved any other file.
- Modified accordion to gave the padding in the body so the box-shadow is not clipped.
- Added a new story showing cards in an accordion.

**Potential Technical Debt:**
The body container will now animate up while the accordion is being closed. This had to be done to make sure the padding between the header and body is correct.

Fixes chec/dashboard#183

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,1 +1,2 @@
 dist/*
+src/stories/index.stories.js
--- a/src/components/ChecAccordion.vue
+++ b/src/components/ChecAccordion.vue
@@ -59,10 +59,10 @@ export default {
 
 <style lang="scss">
 .accordion {
-  @apply rounded bg-gray-100 p-4;
+  @apply rounded bg-gray-100;
 
   &__heading {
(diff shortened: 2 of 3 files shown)


robbieaverill:
The deploy preview looks broken: https://deploy-preview-246--chec-ui.netlify.app/?path=/story/components-accordion--with-card

jaepass:
should down arrow toggle duration be changed here to 200 too? https://github.com/chec/ui-library/pull/248#issuecomment-662165836

john-raymon:
The 200 duration seems a bit slow, now that I'm looking at in action with the accordion. What do you think?

ScopeyNZ:
Nope it feels good to me.

PR #248, opened 2020-07-21
[CHEC-716] Update dropdown icon to flip on active

jaepass:
- Need to clarify if this task is only to update to ChecIcon and flip on active. Looks like the popover option menu was changed to the ellipsis icon so I didn't apply to that
- Was a bit confused on BEM structure here, not sure if i applied the open state correctly on the svg as if you click right on the svg path, dropdown doesn't toggle. But if on dropdown control field and icon area, dropdown does toggle

### DIFF
--- a/src/components/BaseDropdown.vue
+++ b/src/components/BaseDropdown.vue
@@ -358,8 +358,16 @@ export default {
     width: inherit;
 
     &__label {
-      @apply text-left absolute inline-block origin-top-left transition-transform duration-150
+      @apply
+        text-left
+        absolute
+        inline-block
+        origin-top-left
+        transition-transform
+        duration-150
+        cursor-pointer
         truncate w-10/12;
+
       backface-visibility: hidden;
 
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
> * Need to clarify if this task is only to update to ChecIcon and flip on active. Looks like the popover option menu was changed to the ellipsis icon so I didn't apply to that

Yep. Popovers are ellipsis's now, which is a more recent change

> * Was a bit confused on BEM structure here, not sure if i applied the open state correctly on the svg as if you click right on the svg path, dropdown doesn't toggle. But if on dropdown control field and icon area, dropdown does toggle

You've done right. The `--modifier` allows you to target a containing element (and increase specificity) under BEM. Really we're breaking the rules a bit by selecting `svg` specifically, but 🤷‍♂️ .

-----

Had a quick play. Looking pretty good. I showed @dvnkshl and we think the duration should be a bit shorter. Maybe `duration-200`? While you're there, can you ensure that the whole selector has the `cursor-pointer` class too?

Issue #169, opened 2020-06-04
ChecIcon should have a "size" prop with various options in it, also free text option


robbieaverill:
To use an icon at the moment you must provide `h-4 w-4` as classes to give it a size.

We should make this a prop with size options like xs, sm, md, lg etc, and also allow null with custom sizes passed in like they are currently.

PR #249, opened 2020-07-22
Add icon size prop to ChecIcon in line with font size

ScopeyNZ:
I've been missing something like this every time I've used an icon inline with text, so here's an option to specify the size as a prop. This adds a tailwind plugin that generates `icon-[size]` tailwind "utilities" to set the width/height to match the font size.

### DIFF
--- a/tailwind-plugins/iconSize.js
+++ b/tailwind-plugins/iconSize.js
@@ -1,0 +1,14 @@
+const plugin = require('tailwindcss/plugin');
+
+module.exports = plugin(({ addUtilities, config }) => {
+  const fontSizes = config('theme.fontSize');
+
+  // Reduce the existing font sizes and make a height/width definition for each one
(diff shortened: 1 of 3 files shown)


robbieaverill:
https://github.com/chec/ui-library/issues/169

tailwind-plugins/iconSize.js
```diff
@@ -0,0 +1,14 @@
+const plugin = require('tailwindcss/plugin');
+
+module.exports = plugin(({ addUtilities, config }) => {
+  const fontSizes = config('theme.fontSize');
+
+  // Reduce the existing font sizes and make a height/width definition for each one
+  addUtilities(Object.entries(fontSizes).reduce((acc, [label, size]) => ({
```
robbieaverill, line 7: Is this so you avoid passing `h-4 w-4` (for example)?

PR #250, opened 2020-07-22
[CHEC-787] - Fix bug where Dropdown options (popover) list goes off viewport screen

### LINKED ISSUE #211: Dropdown options list goes off screen
image

Ideally it has a max-height, and won't go past the bounds of the viewport. Also, check with the design/UX if it should pop-up if there's more space above.

john-raymon:
Essentially makes the popover/options list have a `max-height` equal to the bottom px distance of the dropdown-el relative to bottom of the window view-port, thus not allowing it to the exceed the viewport screen. (utilizes https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect)

fixes #211

### DIFF
--- a/src/components/BaseDropdown.vue
+++ b/src/components/BaseDropdown.vue
@@ -36,7 +36,14 @@
         <ChecIcon icon="down" />
       </div>
     </div>
-    <BasePopover v-show="showDropdown" class="dropdown__base-popover" :style="{ width: `${dropdownElWidth}px` }">
+    <BasePopover
+      v-show="showDropdown"
+      ref="popper-el"
+      class="dropdown__base-popover"
+      :style="{
+        width: `${dropdownElWidth}px`,
+      }"
+    >
       <BaseOption
         v-for="option in renderableOptions"
         :key="option.value"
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Ok so it looks like we're not using `popper.js` for this. We probably should be, because it probably would've helped you with this bug.

I don't think we can accept this as is, because it's very broken when the dropdown appears at the bottom of the viewport - the dropdown has no height. In this case it should drop "up".

Do you think it would be hard to change this to use popper instead of doing whatever we're doing now?

john-raymon:
@ScopeyNZ eh, I think that essentially means installing Popover.js in this component instead of just fixing it by following the existing logic that exist with set `width: `${dropdownElWidth}px`, but I like the idea of Popover.js in here also.

ScopeyNZ:
We already use popper for tooltips, so wouldn't be any extra bundle size. You'll get more compatibility by using popper though, like drop-ups and what have you.

john-raymon:
Good points, thanks! Adding popper.js now

john-raymon:
Pushed up commit adding popper.js.

ezgif-4-f41aa1b018f5

PR #251, opened 2020-07-23
[CHEC-793] Add new navigation

ScopeyNZ:
This is based on top of #249 (icon sizes). I can rebase that away if it becomes blocking.

This also updates the anti-aliasing and font-size weirdness we've always had, and updates BaseButton to use `router-link` if `tagtype === 'route'`


ScopeyNZ:
Breaking changes to dashboard are fixed in chec/dashboard#226. I'll update that PR with a `yarn upgrade` when it's merged.

robbieaverill:
#249 is merged

robbieaverill:
From a UX perspective:

* I find it strange that hovering over the Chec logo does nothing. I'd expect it to be a link to the home screen, and also open the nav
* I find it strange that you have to hover over a nav icon to expand the menu. I realise it's weird to make assumptions about which to show, but I think we should open the menu when you hover over any of the menu, and I think that it should always have the active section (e.g. orders or products) active unless you've hovered over a specific nav item
* The current user section looks weird, probably just from the filler content that's there, but it's probably worth adding a mockup of what will actually live there so we know it will look good in dash when deployed there
* It really bothers me that our nav isn't keyboard accessible, but don't let that block this PR
* There's no hover state on the menu links
* Is categories becoming a sub-item of products, or is that just an example? Most of these menu sections don't need anything other than their title, so having the nested menu seems overkill IMO

Other things:
image

Non blocking suggestion: This could benefit from a small description of what it is, maybe as a subtitle under "Spaces storefront"

src/components/BaseButton.vue
```diff
@@ -11,7 +11,7 @@ export default {
     tagType: {
       type: String,
       validate(type) {
-        return ['link', 'button'].includes(type);
+        return ['link', 'button', 'route'].includes(type);
```
robbieaverill, line 14: Can you update the docblock plz

ScopeyNZ:
I think I missed the hover state on the links - I'll add that. I'll ping @blazepollard in Slack so he sees that feedback.

ScopeyNZ:
I've updated the PR with:

- The designed hover state for links (I've done this globally so the whole UI has it now)
- A better example for the footer in the nav
- Adjusted doc-blocks.

I don't think I can address any of that UX feedback in this PR, but I've passed it onto Blaze.

I should probably check the impact of my link styling - I'll do that now.

ScopeyNZ:
> I should probably check the impact of my link styling - I'll do that now.

Looks good.

jaepass:
does this need to be rebased first?

ScopeyNZ:
Thanks for the reminder. I've done that now 🙂

robbieaverill:
Follow up:

* Can you make the chec logo linkable to the root route? Blaze: "The Chec logo could redirect to the dashboard home view when clicked" https://madebychec.slack.com/archives/C0259UZ4Q/p1595527395001400?thread_ts=1595523012.000900&cid=C0259UZ4Q
* When I hover over my name, should it switch to the "my account" context? It does if you move your mouse left over the top of the avatar, but doesn't if you only hover the name

Looks pretty good otherwise 👍 

Non-blocking things I'm going to mention anyway:

* "It really bothers me that our nav isn't keyboard accessible, but don't let that block this PR" -> still true
* "There's no hover state on the menu links" -> basically still true, but that's a design issue (gray 600 to 500)

ScopeyNZ:
I've addressed the two bits of feedback you gave around logo clicking and username hovering. And I also added in keyboard navigation as a treat.

I haven't addressed link hover states. Not sure if Blaze will decide to do something custom for the nav, or do something globally to all links. I've reminded Blaze again though and it's on his radar 🙂

ScopeyNZ:
Using the "pop-out" for testing might be easier here (top right):

Screen Shot 2020-07-29 at 5 13 21 PM

src/components/ChecNavigation.vue
```diff
@@ -1,162 +1,494 @@
...
 
 <style lang="scss">
 .chec-navigation {
-  @apply fixed top-0 left-0 w-20 h-screen shadow-sm p-4 flex
-    flex-col transition-all duration-300 ease-in-out overflow-auto;
+  @apply fixed top-0 left-0 h-screen z-20;
+
+  &__skip {
+    @apply fixed p-1 bg-white;
+    left: -500px;
```
robbieaverill, line 419: Magic number

robbieaverill:
The keyboard navigation feels weird that I have to press the right key to go into sub-menus, I'd expect tab to take me into the menu then out of it at the end. We don't need to make accessibility a blocker at the moment anyway, and it's better than not having any so thanks for adding it.

ScopeyNZ:
From my (admittedly brief) research on this, the recommendation was that complex navs should use arrow keys, and still allow the user to tab through high level nav quickly. I was conscious of the fact that when we have many sub-links, just getting to the "settings" category would take like 20 tab presses.

Issue #158, opened 2020-05-29
visual bug when <text-field> is given a value but no label


john-raymon:
- visual bug when <text-field> is given a value but no label, padding off-set within inner input still renders
![image](https://user-images.githubusercontent.com/30238579/83290526-bd47b580-a1b4-11ea-8fc0-3664170cec2a.png)

robbieaverill:
@john-raymon you still working on this?

john-raymon:
@robbieaverill No I'm not, slipped away. I have to check if this bug still occurs.

robbieaverill:
It does, I saw it recently

robbieaverill:
Fixed in https://github.com/chec/ui-library/pull/257

PR #252, opened 2020-07-24
Refactor label to not cover half the input

ScopeyNZ:
The problem was that the label was actually covering half the input, and we were using a `:before` psuedo-element to actually render the text. I couldn't figure out why we did this, so I've refactored it to use the actual label, and made it only as big as it's contents. I also fixed up the alignment so that it's correctly aligned with the text field value.

Fixes chec/dashboard#53

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -19,7 +19,7 @@
       :data-content="label"
       :for="id"
     >
-      <span class="invisible">{{ label }}</span>
+      {{ label }}
     </label>
     <div v-if="$slots.default" ref="rightContentSlot" class="text-field__right-content">
       <slot />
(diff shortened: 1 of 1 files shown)


robbieaverill:
Does this mean we don't need `{force: true}` in Cypress anymore?

robbieaverill:
Multiline is funky
image

robbieaverill:
Otherwise looking good

john-raymon:
Nice, this doesn't https://github.com/chec/ui-library/issues/158#issuecomment-663280963 right

ScopeyNZ:
Nope it doesn't solve that issue. That'd require some JS changes I assume.

john-raymon:
Changing `margin: 1px;` to `padding: 1px;` on line 328 seems to fix the funky multiline bug. https://github.com/chec/ui-library/blob/f63619e2a0befdcc115f8e93bded180b7a674463/src/components/TextField.vue#L328

ScopeyNZ:
There was more to it than that. The underlay was broken.

ScopeyNZ:
It should fix the `{force:true}` hack in dash yeah.

PR #255, opened 2020-07-24
Simplify dropdown DOM an make styling consistent with text fields

### LINKED ISSUE #228: Dropdown height is not the same as text field height
image

ScopeyNZ:
Fixes #228

### DIFF
--- a/src/stories/FormFields.stories.mdx
+++ b/src/stories/FormFields.stories.mdx
@@ -31,7 +31,7 @@ import ChecDropdown from '../components/ChecDropdown.vue';
           <div class="max-w-3xl w-full mx-auto grid grid-cols-2 gap-1">
             <TextField
               name="input-name"
-              label="Dropdown"
+              label="Text field"
               placeholder="Example Placeholder"
               v-model="inputValue"
             />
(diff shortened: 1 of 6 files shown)


src/stories/FormFields.stories.mdx
```diff
@@ -0,0 +1,47 @@
+import { Meta, Props, Story, Preview } from '@storybook/addon-docs/blocks';
+import { action } from "@storybook/addon-actions";
+import { withKnobs, object, boolean, text, select } from '@storybook/addon-knobs';
+import { withInfo } from 'storybook-addon-vue-info'
+import TextField from '../components/TextField.vue';
+import ChecDropdown from '../components/ChecDropdown.vue';
+
+<Meta title="Form fields" component={TextField} />
```
robbieaverill, line 8: Did you deliberately put this into its own "Overview" story? If so, all good.

The text field's label says "dropdown" which confused me.

ScopeyNZ: Yep I did. I've renamed the field name. I forgot to change it back after I was making sure the fields were consistent.

ScopeyNZ:
Thanks. My simplification moved the click handler, which meant that clicks on the popover were causing it to toggle. I've fixed it by using a portal to hoist the popper to the body - this also fixes an issue where the popper wouldn't break out of overflow containers: https://chec-ui.netlify.app/?path=/story/form-fields-dropdown--within-a-chec-modal-popping-out-of-overflow .

src/stories/FormFields.stories.mdx
robbieaverill: The text field label should still be updated

PR #263, opened 2020-07-28
add properties to package.json for npm publishing

john-raymon:


package.json
```diff
@@ -1,8 +1,13 @@
...
   "name": "@chec/ui-library",
-  "version": "0.1.0",
-  "private": true,
+  "version": "0.0.1",
   "main": "dist/ui-library.common.js",
+  "homepage": "https://github.com/chec/ui-library",
+  "author": "Chec Platform Inc",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/chec/ui-library.git"
```
robbieaverill, line 9: This differs from the [CLI](https://github.com/chec/cli/blob/master/package.json#L77), [commerce.js](https://github.com/chec/commerce.js/blob/master/package.json#L15-L18), and [seeder](https://github.com/chec/seeder/blob/master/package.json#L9) configs. May I suggest you use `"repository": "chec/ui-library"` to keep it simple?

PR #265, opened 2020-07-30
[CHEC-825] Add loading prop to ChecOption and to ChecDropdown

ScopeyNZ:
Part 2 of 3 for this issue. Adds a loading indicator to dropdown option lists, and a way to toggle it with a prop

### DIFF
--- a/src/components/ChecOption.vue
+++ b/src/components/ChecOption.vue
@@ -7,22 +7,30 @@
     @click="emitOptionsSelectedEvent"
     @keyup.enter="emitOptionsSelectedEvent"
   >
-    <input
-      v-if="showCheckbox"
-      :id="option.value"
-      type="checkbox"
-      class="mr-3"
-      :checked="checked"
-      :disabled="option.disabled"
-      :indeterminate.prop="indeterminate"
-    >
-    <slot />
+    <div v-if="loading" class="option__loading-container">
+      <ChecLoading />
+    </div>
+    <template v-else>
+      <input
+        v-if="showCheckbox"
+        :id="option.value"
+        type="checkbox"
+        class="mr-3"
+        :checked="checked"
+        :disabled="option.disabled"
+        :indeterminate.prop="indeterminate"
+      >
+      <slot />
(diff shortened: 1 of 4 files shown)


src/components/ChecOption.vue
```diff
@@ -89,19 +102,25 @@ export default {
 .option {
   @apply w-full flex items-center text-sm px-4 py-3 outline-none cursor-pointer bg-white;
 
-  &:hover {
+  &:hover:not(.option--loading) {
     @apply bg-gray-100;
   }
```
john-raymon, line 107:
```suggestion
  &:hover {
    &:not(.option--loading) {
      @apply bg-gray-100;
    }
  }
```

ScopeyNZ: Why?

src/stories/components/ChecDropdown.stories.mdx
```diff
@@ -44,10 +44,13 @@ Todo list
...
         errored: {
-          default: boolean('Errored', false)
+          default: boolean('Errored', false),
         },
         disabled: {
-          default: boolean('Disabled', false)
+          default: boolean('Disabled', false),
+        },
+        loading: {
+          default: boolean('Loading', false),
```
john-raymon, line 53: What do you think of having the loading option first on the list and initially loading by default?

src/components/ChecOption.vue
john-raymon: Thought id suggest since we're doing it within the :active, and :focus states, although we have separate styles within those rules. Can ignore it though 🙈

src/stories/components/ChecDropdown.stories.mdx
ScopeyNZ: The loading indicator appears on the bottom of the list in the designs, and I think it makes sense for a "continuous scroll" use-case. If you're talking about first on the knobs list - I'm not sure that it's really that important.

robbieaverill:
Needs a rebase

PR #268, opened 2020-08-03
[CHEC-827] - create chec-file-uploader, using dropzone

john-raymon:
- also creates primitive component file-row for chec-file-uploader component to use
ezgif-5-aeb62a1abd73


src/components/ChecFileUploader/FileRow.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+  <div class="chec-file-row" :class="{ 'chec-file-row--error': hasError }">
+    <div class="chec-file-row__left">
+      <svg
```
ScopeyNZ, line 4: Don't we have `<ChecLoading>` for this?

src/components/ChecFileUploader/FileRow.vue
```diff
@@ -0,0 +1,158 @@
...
+    ChecIcon,
+  },
+  props: {
+    loading: {
+      type: Boolean,
+    },
+    error: {
+      type: Boolean,
+    },
+    file: {
```
ScopeyNZ, line 78: What do you think about being more verbose in the props here. Rather than taking in `file` - you could take in `filename`, `mimetype` and `size`?

src/components/ChecFileUploader/FileRow.vue
```diff
@@ -0,0 +1,158 @@
...
+      <p class="chec-file-row__label">
+        {{ file.upload && file.upload.filename || '--' }}
+      </p>
+    </div>
+    <div class="chec-file-row__right">
+      <p v-if="!hasError" class="chec-file-row__type">
+        {{ file.type }}
+      </p>
+      <p v-if="!hasError" class="chec-file-row__size">
+        {{ size }}MB
```
ScopeyNZ, line 52: https://www.npmjs.com/package/filesize

Then we can support smaller/larger files with a slightly nicer interface.

src/components/ChecFileUploader/FileRow.vue
```diff
@@ -0,0 +1,158 @@
...
+    @apply text-white text-xs;
+  }
+
+  &__type {
+    @apply text-white text-xxs uppercase mr-1;
+    letter-spacing: 1px;
+  }
+
+  &__size {
+    @apply text-white text-xxs uppercase mr-1;
```
ScopeyNZ, line 136:
```suggestion
    @apply text-white caps-xxs mr-1;
```

src/components/ChecFileUploader/FileRow.vue
```diff
@@ -0,0 +1,158 @@
...
+      @apply opacity-75 outline-none;
+    }
+  }
+
+  &__label {
+    @apply text-white text-xs;
+  }
+
+  &__type {
+    @apply text-white text-xxs uppercase mr-1;
```
ScopeyNZ, line 131:
```suggestion
    @apply text-white caps-xxs mr-1;
```

src/components/ChecFileUploader/FileRow.vue
john-raymon: Ah I think we do, but it's currently not the same. Do you think it's better to update chec-loading to support a secondary style?

ScopeyNZ: I think you should use the component that was built for that purpose, and we can update the style later.

robbieaverill: Using common patterns for things like this forces us to be consistent with our design implementation

src/components/ChecFileUploader.vue
```diff
@@ -0,0 +1,123 @@
...
+      delete files[file.upload.uuid];
+      this.files = { ...files };
+      this.$emit('file-removed', file);
+    },
+  },
+};
+</script>
+
+<style lang="scss">
+.chec-dropzone {
```
robbieaverill, line 100: Can you name these classes `.chec-file-uploader` rather than dropzone/the underlying library?

src/components/ChecFileUploader.vue
```diff
@@ -0,0 +1,123 @@
...
+
+  &__inner {
+    @apply relative border border-dashed border-gray-400 rounded m-2 p-8;
+  }
+
+  .button {
+    @apply mx-auto;
+  }
+
+  input {
```
robbieaverill, line 111: Is there something else we can use to target this? The linter will complain about this being too specific

src/components/ChecLoading.vue
```diff
@@ -67,6 +102,10 @@ export default {
   }
 }
 
+.secondary-animation {
+  @apply animate-spin h-full w-full;
+}
```
robbieaverill, line 107: I disagree with this approach. We should use the same loading component everywhere. If there are new loading components we can add them as new components - making a component like this work in two ways is an anti-pattern. You're better off making a parent component like `ChecLoading` then creating `Loading` and `FileLoading` components that extend them and using those.

All I've seen design-wise is your animated GIF in the PR description. It looks like it's a regular `ChecLoading` component but in a different colour. Is that right? If so, maybe we can add a `variant` prop and accept some sort of "inverted" option where we switch the colour?

This could also be separated into a different issue and pull request too.

robbieaverill: Implementation wise it appears that it's colour and also the direction the spinner moves in. My recommendation is that we ask Blaze to use a consistent loading indicator, and give it a variant that works on a dark background.

john-raymon: Thank you Robbie. I'll make a separate PR introducing the loading variation in the form of components extending chec-loading. 

>All I've seen design-wise is your animated GIF in the PR description. It looks like it's a regular ChecLoading component but in a different colour. Is that right? If so, maybe we can add a variant prop and accept some sort of "inverted" option where we switch the colour?

Yes you're right, it also renders a white overlay that's not prop configurable. Should the variant prop handle that also?

ScopeyNZ:
How's this going @john-raymon? Just want to make sure you saw my comment about the missed A/C?

john-raymon:
@ScopeyNZ Yes, wrapping this up was separating the chec-loading work to it's own PR https://github.com/chec/ui-library/pull/271

ScopeyNZ:
@john-raymon if you can add doc-blocks to props and events, and remove yourself as an assignee when you've finished working on this, I'll re-review. I think that'll help keep things cleaner rather than littering it with reviews while it's WIP.

Let me know if you want any help working on this.

john-raymon:
@ScopeyNZ Added the doc-block comments and some other necessary changes such as the error file-row state and more. All ready to review :)

src/components/ChecFileUploader.vue
```diff
@@ -0,0 +1,152 @@
...
+
+  &__input {
+    display: none;
+  }
+
+  .file-rows-container {
+    @apply w-full flex flex-col mb-2;
+
+    .chec-file-row {
+      @apply mb-2;
```
ScopeyNZ, line 148: We've been using `space-y-2` for this in other places. We currently have to apply it with a classname (not `@apply`), as Tailwind doesn't support adding that helper with `@apply` yet.

ScopeyNZ:
Oh also the `FileRow` story wasn't working for me.

PR #269, opened 2020-08-05
Add "variants" to accordions, allowing different toggle types

ScopeyNZ:
I've also:

- tried to smooth out the animations a bit (again) - although it's still a bit janky if you spam the button. I've checked that the cards in accordions use-case is unaffected.
- added the button label to the "title" attribute on the existing toggle button - hopefully improving accessibility
- Added an option to prefix the label on a switch, which is coming up a bit in new designs

This is out of cycle work. I'm supposed to be scoping the add product work. I can't help getting a bit bored though :slightly_smiling_face:. The new add product designs use this pattern (https://www.figma.com/file/og79WKi1zrrBgZWayZY4ik/Dashboard-Web?node-id=4196%3A52776). The accordion designs in Figma don't show this option, but the component is an accordion in Figma, so I assume that Blaze expected it would be an accordion variant.

### DIFF
--- a/src/components/ChecAccordion.vue
+++ b/src/components/ChecAccordion.vue
@@ -85,10 +85,14 @@ export default {
   },
   computed: {
     resolvedButtonLabel() {
-      return this.buttonLabel || (this.variant === 'switch'
-        ? this.$t('accordion.switchLabel')
-        : this.$t('accordion.toggleLabel')
-      );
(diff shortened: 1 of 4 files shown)


ScopeyNZ:
Annoying. I thought I checked that. I've pushed a change.

PR #272, opened 2020-08-13
[CHEC-841] Add required prop to text field component

jaepass:
- Added required prop for toggling native 'required' attribute and targeted styling
*Went with the first option in the task, hope I understood the a/c's correctly*

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -30,7 +30,12 @@
       :for="id"
     >
       {{ label }}
-      <span v-if="required">(required)</span>
+      <span
+        v-if="required"
+        class="text-field__required-text"
+      >
+        (required)
+      </span>
     </label>
     <div v-if="$slots.default" ref="rightContentSlot" class="text-field__right-content">
       <slot />
@@ -411,7 +416,7 @@ export default {
   }
 
   &--required {
-    .text-field__label span {
+    .text-field__required-text {
       @apply text-gray-400;
     }
   }
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -16,6 +16,7 @@
     />
     <input
       v-else
+      :required="required"
       v-bind="sharedInputProps"
       :type="$attrs.type || 'text'"
       :style="{ 'padding-right': hasSlot && `${slotWidth}px` }"
@@ -29,6 +30,7 @@
       :for="id"
     >
       {{ label }}
+      <span v-if="required">(required)</span>
     </label>
     <div v-if="$slots.default" ref="rightContentSlot" class="text-field__right-content">
       <slot />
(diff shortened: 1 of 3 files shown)


src/components/TextField.vue
```diff
@@ -29,6 +30,7 @@
       :for="id"
     >
       {{ label }}
+      <span v-if="required">(required)</span>
```
ScopeyNZ, line 33:
```suggestion
      <span v-if="required" class="text-field__required-text">(required)</span>
```

src/components/TextField.vue
```diff
@@ -400,6 +410,12 @@ export default {
     }
   }
 
+  &--required {
+    .text-field__label span {
```
ScopeyNZ, line 414:
```suggestion
    .text-field__required-text {
```

ScopeyNZ:
Looking really good. Just got that suggestion about BEM, but otherwise LGTM.

src/components/TextField.vue
ScopeyNZ: FYI the specificity of what you've got is 3, one from the text-field--required, one from `text-field__label` and another from `span`.

BEM says you should always have specificity of 1, unless there's a modifier involved, where it can be 2. `&--required` is the modifier here.

src/components/TextField.vue
jaepass: ok got it thanks Guy, im going to look for a crash course 101 on BEM, need to nail this down!

src/components/TextField.vue
```diff
@@ -400,6 +415,12 @@ export default {
     }
   }
 
+  &--required {
+    .text-field__required-text {
```
robbieaverill, line 419: Personally I don't think you need to nest this inside `&--required` at all. If the element is there, the styles for it should be there too. If it's not there then it doesn't matter. The parent modifier class doesn't seem particularly relevant to me.

src/components/TextField.vue
```diff
@@ -29,6 +30,12 @@
       :for="id"
     >
       {{ label }}
+      <span
+        v-if="required"
+        class="text-field__required-text"
+      >
+        (required)
```
robbieaverill, line 37: Would you mind making this translatable? This would mean adding something like this to `lang/en.js`:

```js
{
  "general": {
    "required": "(required)"
  }
}
```

Then do this:
```suggestion
        {{ $t('general.required') }}
```

src/components/TextField.vue
jaepass: got it thanks will remove nesting but leave modifier?

ScopeyNZ: I think he's saying you can remove the modifier. It can just be `.text-field__required-text` because the "required text" style is applicable regardless of whether the field is actually required or not.

robbieaverill: The modifier might be useful at some point, I wasn't explicitly saying you should remove that, just the nesting of this definition

PR #275, opened 2020-08-18
Added and edited colors

jaepass:
Docs design uses gray-700 which is a new color so I've added that along with other new color changes while I'm at it


jaepass:
no primary-gradient-2 isn't used anywhere i've check all projects

PR #278, opened 2020-08-21
[CHEC-848] Update tags to match new designs in Figma

robbieaverill:
Also adds a new button state for "tag" which is just for tag groups


robbieaverill:
Deploy preview: https://5f3f18089ad4700007f8ff2a--chec-ui.netlify.app/

PR #282, opened 2020-08-25
[CHEC-842] Feature/vertical nav

jaepass:
- Added vertical navigation UI component
- I was running into issues with this early on as I was trying to use options as props and wasn't able to nail down how to detect the active item correctly. With suggestions from Robbie and John, refactored to use Vue router's router link component which has its native active class
- Having trouble trying to figure out alternatively how to programmatically push the routes https://router.vuejs.org/guide/essentials/navigation.html#programmatic-navigation in order to add keyboard events (I have not added keyboard events) <- would love some help on this if this is part of a/cs

### DIFF
--- a/src/components/ChecVerticalNavItem.vue
+++ b/src/components/ChecVerticalNavItem.vue
@@ -1,0 +1,72 @@
+<template>
+  <li
+    class="vertical-navigation-item"
+    :class="{ 'vertical-navigation-item--active': active }"
+    @keyup.down="focusNextItem"
+    @keyup.up="focusPreviousItem"
+    @click="handleClick"
+  >
+    <a :href="link">
+      <!--
+        @slot Label to display as the navigation item label
+      -->
+      <slot v-bind="{ active, disabled }" />
+    </a>
+  </li>
+</template>
+
+<script>
+export default {
+  name: 'ChecVerticalNavItem',
+  props: {
+    /**
+     * Determines the link of the navigation item
+     */
+    link: {
+      type: String,
+      required: true,
+    },
+    /**
+     * If this navigation item should appear selected (active)
+     */
+    active: Boolean,
+    disabled: Boolean,
+  },
+  methods: {
+    /**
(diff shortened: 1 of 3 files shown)


src/components/ChecVerticalNavItem.vue
```diff
@@ -0,0 +1,78 @@
...
+      return this.$route.path;
+    },
+  },
+};
+</script>
+
+<style lang="scss">
+.vertical-navigation {
+  &__item {
+    @apply px-2 py-3 border-gray-200 border-b rounded cursor-pointer;
```
ScopeyNZ, line 58:
```suggestion
    @apply px-2 py-3 border-gray-200 rounded cursor-pointer text-sm leading-tight;
```

src/components/ChecVerticalNavigation.vue
```diff
@@ -0,0 +1,29 @@
...
+<script>
+
+export default {
+  name: 'ChecVerticalNavigation',
+};
+</script>
+
+<style lang="scss">
+.vertical-navigation {
+  @apply rounded shadow-sm w-56 bg-white px-3 py-5;
```
ScopeyNZ, line 23:
```suggestion
  @apply rounded-md shadow-sm bg-white px-3 py-5;
```

The removal of the `w-56` is something I think we should try to get in the habit of doing. The implementor should be able to choose width I think.

src/components/ChecVerticalNavItem.vue
```diff
@@ -0,0 +1,78 @@
...
+
```
ScopeyNZ, line 59:
```suggestion
    & + & {
        @apply border-t;
    }

```
This should get the border between (and not on the last item). Might need to test this before you accept it though.

jaepass:
Thanks Guy will edit stat!

src/components/ChecVerticalNavItem.vue
jaepass: this worked thanks!

PR #283, opened 2020-08-26
[CHEC-861] Updating Input Hover, Active and dark versions

kvisca:
**Changes:**   
- Adding in the Hover and Active updates
- Adding in the dark version for text fields

**Note:**
I'm dictating it through a new prop called `styleVariant`.  The reasoning for this is the error and disabled states are considered variants. I think we may want to take a look into how this component is structured in the future as a state is probably not a variant. I left variant untouched so as to not cause issues with integrated code. 

**Screenshots:**

Screen Shot 2020-08-26 at 2 10 43 PM
Screen Shot 2020-08-26 at 2 10 33 PM

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -325,15 +325,18 @@ export default {
       @apply transition duration-150 border-gray-400;
     }
   }
-  &--dark{
+
+  &--dark {
     .text-field__label {
       @apply text-gray-300;
     }
+
     .text-field__input {
       @apply
         text-white
         border-gray-600
         bg-gray-600;
+
       &::placeholder {
         color: rgba(0, 0, 0, 0);
       }
@@ -356,6 +359,7 @@ export default {
       }
     }
   }
+
   &__right-content {
     // Alignment & spacing
     @apply absolute flex items-center h-full right-0 top-0 mx-4;
@@ -407,7 +411,7 @@ export default {
 
     &:focus,
     &:active {
-      box-shadow: 0px 0px 0px 4px rgb(231, 126, 143, 0.3);
+      box-shadow: 0 0 0 4px rgb(231, 126, 143, 0.3);
     }
 
   }
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -121,6 +121,16 @@ export default {
     required: {
       type: Boolean,
     },
+    /**
+     * The style variant for the text Input. One of "light" or "dark"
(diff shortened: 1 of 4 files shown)


src/components/TextField.vue
```diff
@@ -304,11 +318,45 @@ export default {
...
+        color: rgba(0, 0, 0, 0);
+      }
+
+      &:placeholder-shown {
+        @apply py-4;
+
+        + .text-field__label {
+          transform: scale(1, 1);
+        }
+      }
```
ScopeyNZ, line 350:
```suggestion
```

Seems like unnecessary duped code? It felt like this was causing a bug for me locally too.

src/components/TextField.vue
```diff
@@ -360,6 +408,12 @@ export default {
     &:active {
       @apply border-red-300;
     }
+
+    &:focus,
+    &:active {
+      box-shadow: 0 0 0 4px rgb(231, 126, 143, 0.3);
```
ScopeyNZ, line 414: Should this be a tailwind config change too?

tailwind.config.js
```diff
@@ -50,6 +50,8 @@ module.exports = {
       sm: '0px 0px 1px rgba(40, 51, 65, 0.2), 0px 1px 4px rgba(40, 51, 65, 0.1)',
       md: '0px 3px 12px rgba(40, 51, 65, 0.1), 0px 0px 1px rgba(40, 51, 65, 0.2)',
       lg: '0px 4px 16px rgba(40, 51, 65, 0.1), 0px 0px 1px rgba(40, 51, 65, 0.2)',
+      light: '0px 0px 0px 4px #D6E4F7',
+      dark: '0px 0px 0px 4px #41546C',
```
ScopeyNZ, line 54: I wonder if we could improve the naming here. Looking at the options in the tailwind config doesn't really immediately look like "input focus state shadow". Not quite sure what this could be... Maybe just `light-focus` and `dark-focus`?

src/components/TextField.vue
kvisca: You are correct, removed

tailwind.config.js
kvisca: Updated based on your suggestion

src/components/TextField.vue
kvisca: Added it to the  tailwind config as `error-focus`

PR #284, opened 2020-09-08
[CHEC-915]  Modify textfield to have currency type

kvisca:
Added a currency prop that shows a dollar sign before the input. Value will still be unaffected. Down the line, we could make it so the symbol is definable programmatically but for now, I have included it as a prop. 

Screen Shot 2020-09-08 at 6 15 56 PM

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -96,11 +96,11 @@ export default {
     */
     multiline: Boolean,
     /**
-    * Display currency text field
+    * Display text field in currency mode
     */
     currency: Boolean,
     /**
-    * Currency symbol
+    * Currency symbol for use in currency mode
     */
     currencySymbol: {
       type: String,
@@ -343,6 +343,7 @@ export default {
     .text-field__label {
       @apply text-gray-300;
     }
+
     .text-field__currency {
       @apply text-white;
     }
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -343,6 +343,9 @@ export default {
     .text-field__label {
       @apply text-gray-300;
     }
+    .text-field__currency {
+      @apply text-white;
+    }
 
     .text-field__input {
       @apply
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -37,6 +37,9 @@
         {{ $t('general.required') }}
       </span>
     </label>
+    <div class="text-field__currency">
+      $
+    </div>
     <div v-if="$slots.default" ref="rightContentSlot" class="text-field__right-content">
       <slot />
     </div>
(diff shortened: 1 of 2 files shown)


src/components/TextField.vue
```diff
@@ -94,6 +97,19 @@ export default {
     multiline: {
       type: Boolean,
     },
+    /**
+    * Display currency text field
+    */
+    currency: {
+      type: Boolean,
+    },
```
ScopeyNZ, line 105:
```suggestion
    currency: Boolean,
```

src/components/TextField.vue
```diff
@@ -456,16 +489,31 @@ export default {
...
+  }
+
   &:not(.text-field--disabled) .text-field__input {
     &:focus,
     &:active {
       + .text-field__label {
         @extend %filled-transformation;
       }
+
+      ~ .text-field__currency {
```
ScopeyNZ, line 514: Do you need this "direct descendant" selector (or whatever it's called)?

ScopeyNZ (commented):
Nice! Looking really good. We need to update the colour in dark mode though, it's a little too hard to see.

Otherwise just a minor suggestion and question.

robbieaverill (commented):
Is there a reason you chose to extend the TextField component rather than create a new CurrencyField wrapper around it? Ultimately it doesn't really matter, but I'd like to avoid bloating our existing components where possible. Note that we did extend TextField for the PasswordField rather than adding to TextField (although the initial PR for it did what you've done here as well, and I moaned about it then too)

src/components/TextField.vue
kvisca: Yes, we need it in this case.

src/components/TextField.vue
kvisca: updated for required and multiline as well.

kvisca:
> Is there a reason you chose to extend the TextField component rather than create a new CurrencyField wrapper around it? Ultimately it doesn't really matter, but I'd like to avoid bloating our existing components where possible. Note that we did extend TextField for the PasswordField rather than adding to TextField (although the initial PR for it did what you've done here as well, and I moaned about it then too)

I extended it because that's what the linear task requested.

robbieaverill:
Okay, maybe we need to more explicit in our tasks in future, because the title is "Add a currency component"

ScopeyNZ:
I didn't really intend the description of the Linear task to be taken too literally there. But ultimately I think it would've been more trouble to wrap the text field component anyway.

ScopeyNZ:
Also, I think you may have missed my comment about the text colour in dark mode.

kvisca:
> Also, I think you may have missed my comment about the text colour in dark mode.

Completely missed it! I've updated accordingly.

PR #286, opened 2020-09-10
[CHEC-911] Adding input hints, moving icon and hints to wrapper.

kvisca:
- Created a form field wrapper component to help in cleaning up the complexity and reusability of our inputs.
- Moved icon management to the wrapper
- Created a component for the action button/helper text. 
- Updated stories.


src/components/ChecFormFieldActionPrepend.vue
```diff
@@ -0,0 +1,74 @@
...
+  top: -2px;
+
+  a,
+  span {
+    @apply caps-xxs text-gray-500 mt-1;
+  }
+
+  a {
+    @apply text-blue-500 cursor-pointer;
+  }
```
ScopeyNZ, line 57: Please BEMify this.

src/components/TextField.vue
```diff
@@ -458,16 +423,6 @@ export default {
...
-  &--has-icon {
-    .text-field__label {
-      left: 1.5rem;
-    }
-
-    .text-field__input {
-      @apply pl-10;
-    }
-  }
-
```
ScopeyNZ, line 470: There's some style rules around the old action button that you can remove now.

ScopeyNZ (commented):
Glad to see that you didn't go down the route of moving all the shared attributes (between selects and text inputs) to this wrapper. I don't think we should put the effort into that right now.

I'm not sure about the click event introspection... It'd be great if you could add a knob the story so that the capability (of non-clickable help text) is obvious from the story. It's not clear how we have the help text without making it a clickable link.

I think this is the right technical approach now, and we can expand on this form field component going forward, but I'd prefer that implement components based on the designs as they are now, and we can figure out changes later. Namely a separate tooltip and action button. If we stick with your approach, Blaze will need to know he has the limitation of either help text or an action button, never both, which his designs currently cater for. Also, `actionText` is not a suitable prop name in that case.

ScopeyNZ (approved):
I'm happy. Can you prep a PR that updates the existing places we use icons and action text? We've decided to do this in the past because otherwise the next person that needs an updated UI library for something has a bunch of stuff that they didn't change breaking.

ScopeyNZ:
FYI I think you'll have to PR auth here too.

kvisca:
@ScopeyNZ Went through all the projects and it doesn't look like we are using icons in text fields anywhere yet.  The password field used in auth is its own component and I have updated it accordingly, good catch. 

Shout out if you know of anywhere I may have missed that uses an icon, otherwise we good to go.

ScopeyNZ:
Ok sounds good.

PR #287, opened 2020-09-15
[CHEC-899] Add the ability to dynamically issue URLs for ChecFileUploader

ScopeyNZ:
This PR allows you to create an asynchronous function that resolves a URL where a file should be uploaded, and provide that to the `endpoint` prop instead of a constant URL.


ScopeyNZ:
@kvisca, sorry, this might have some overlap with what you're doing. Happy to get on a call and discuss things if you need.

robbieaverill (approved):
LGTM, untested though

Issue #190, opened 2020-06-09
Dropdown should not allow values that don't exist as options


robbieaverill:
If I make a dropdown and give it options Foo, Bar, and Baz, but then I give it the value of "World", it will retain the "World" value in the selected options list permanently. The others will add or remove as selected which is fine, but I have no way of removing "World" from the value of the field.

This is a problem, for example, if the dropdown contains a list of product IDs, and the populated value comes from an existing API call for something. The product ID is then set as a selected value in the dropdown, but if it doesn't exist in the options then I can't de-select it, and my API calls to save a form might complain that that product doesn't exist - at which point I have no ability to fix it as a user.

Example: create a discount, set some product IDs, go and manually change one of them in the DB to a product ID that doesn't belong to your merchant, then try and edit the discount and save it.

PR #294, opened 2020-09-18
ChecDropdown no longer accepts value attributes that do not exist in the options attribute

robbieaverill:
Fixes https://github.com/chec/ui-library/issues/190

### DIFF
--- a/src/components/ChecDropdown.vue
+++ b/src/components/ChecDropdown.vue
@@ -263,6 +263,8 @@ export default {
     if (this.value !== '' && this.value.length !== 0) {
       this.isFocus = true;
     }
+    this.validateValue();
+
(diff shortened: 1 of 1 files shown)


ScopeyNZ (approved):

ScopeyNZ:
On second thought, the way this is written it'll only throw an error when the component is created, and not if you bind in an invalid value after it's mounted...

robbieaverill:
Correct, there's no way to do that

robbieaverill:
Well I suppose there is a way to do that but I feel like it's unlikely to happen

ScopeyNZ:
Yeah the only way is to set up a watch... This is really a tool to help developers not foot-gun, so I'm happy as is.

PR #299, opened 2020-09-22
Add the ability to drag & drop sort images in the image manager

ScopeyNZ:
I've done some stuff to get the drag image to not be too large, and I've run the on-the-fly design of the drop placeholder and the changes to the image fit past Blaze.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,10 @@
     "lint-js:fix": "vue-cli-service lint --fix",
     "lint-scss:fix": "stylelint \"src/**/*.scss\" \"src/**/*.vue\" --fix",
     "build-lib": "vue-cli-service build --target lib --name ui-library src/index.js",
-    "storybook:build": "vue-cli-service storybook:build -c config/storybook",
-    "storybook:serve": "vue-cli-service storybook:serve -p 6006 -c config/storybook"
+    "build": "vue-cli-service storybook:build -c config/storybook",
+    "storybook:build": "yarn build",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -17,8 +17,10 @@
     "lint-js:fix": "vue-cli-service lint --fix",
     "lint-scss:fix": "stylelint \"src/**/*.scss\" \"src/**/*.vue\" --fix",
     "build-lib": "vue-cli-service build --target lib --name ui-library src/index.js",
-    "storybook:build": "vue-cli-service storybook:build -c config/storybook",
-    "storybook:serve": "vue-cli-service storybook:serve -p 6006 -c config/storybook"
+    "build": "vue-cli-service storybook:build -c config/storybook",
+    "storybook:build": "yarn build",
+    "serve": "vue-cli-service storybook:serve -p 6006 -c config/storybook",
```
robbieaverill, line 22: I had this same thought the other day 😄

src/lang/en.js
```diff
@@ -1,5 +1,18 @@
 export default {
   en: {
+    accordion: {
```
robbieaverill, line 3: I'd put general at the top deliberately, but fair enough for making it alphabetical

robbieaverill (approved):
My only qualm with this is that the grab cursor is everywhere other than the actual drag/reorder icon in the centre, where it's a pointer.

Non blocking.

kvisca (approved):
I changed the pointer on the drag icon as I agree with Robbie. Otherwise, LGTM!

PR #302, opened 2020-09-24
Make Tag and TagGroup classes more unique

robbieaverill:
They currently conflict with anything else we use that has "tag" classes applied to it, e.g. Prism syntax highlighting


ScopeyNZ (approved):
LGTM

PR #306, opened 2020-09-28
Dropdown & textfield force minimized label option.

kvisca:
Adding option to allow text fields and dropdowns to have an always minimized label. This is necessary for the shipping options section. 

Screen Shot 2020-09-28 at 3 20 03 PM

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -164,7 +164,7 @@ export default {
         variant,
         minimizedLabel,
       } = this;
-      console.log(minimizedLabel);
+
       return {
         'text-field--currency': currency,
         'text-field--disabled': variant === 'disabled',
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -84,6 +84,10 @@ export default {
     */
     multiline: Boolean,
     /**
+    * Force label to always be in the minimized position.
+    */
+    minimizedLabel: Boolean,
+    /**
     * Display text field in currency mode
     */
     currency: Boolean,
@@ -158,8 +162,9 @@ export default {
         value,
         styleVariant,
         variant,
+        minimizedLabel,
       } = this;
-
+      console.log(minimizedLabel);
       return {
         'text-field--currency': currency,
         'text-field--disabled': variant === 'disabled',
@@ -168,6 +173,7 @@ export default {
         'text-field--inline-label': label,
         'text-field--modified': label ? !!value : false,
         'text-field--multiline': multiline,
+        'text-field--minimized-label': minimizedLabel,
         [`text-field--${styleVariant}`]: styleVariant !== '',
       };
     },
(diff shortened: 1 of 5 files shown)


robbieaverill:
Why is it necessary for the shipping options section? I can't see why from looking at the designs:

image

kvisca:
> Why is it necessary for the shipping options section? I can't see why from looking at the designs:
> 
@robbieaverill I spoke with Blaze and this was what he intended. If you look at that design, the `On Own` and `With others `do not have a value set and they are in the minimized state. 

It's likely to avoid situations where the input looks broken due to the label being too big at that size. See image for example:

Screen Shot 2020-09-28 at 3 33 12 PM

robbieaverill:
They do have a value set - it's a $ sign. Has that been removed from the designs?

kvisca:
> They do have a value set - it's a $ sign. Has that been removed from the designs?

The dollar sign is not a value though, its a pre-fixture. A PR went in last week do address the currency and what happens is the dollar sign is hidden by default. When the user clicks into the field, the label minimizes and the currency symbol fades in.  You can see how that looks if you toggle on the currency option in the text field in Storybook.

https://chec-ui.netlify.app/?path=/story/form-fields-text-field--defaults

robbieaverill:
It sounds to me like that's diverging from the designs. The designs (to me) appear to always show the $ placeholder regardless of whether you've clicked into it.

This PR appears to be fixing a problem that shouldn't exist in the first place. My recommendation would be to re-evaluate the way that support for a $ prefix has been added. In my mind it is something the component renders as part of the value, e.g. `<input :value="computedValue"` and `computedValue() { return `${this.prefix}${this.value}` etc.

What do you think?

kvisca:
> It sounds to me like that's diverging from the designs. The designs (to me) appear to always show the $ placeholder regardless of whether you've clicked into it.
> 
> This PR appears to be fixing a problem that shouldn't exist in the first place. My recommendation would be to re-evaluate the way that support for a $ prefix has been added. In my mind it is something the component renders as part of the value, e.g. `<input :value="computedValue"` and `computedValue() { return `${this.prefix}${this.value}` etc.
> 
> What do you think?

@robbieaverill I reconfirmed with Blaze and everything is kosher. You should be able to have an input where the label is minimized when no value or currency symbol is present. You should also be able to have a currency field where, by default, the currency is hidden and the label is large in the center. 

Hopefully, that makes sense. Let me know if you want to hop on a call and I can try to explain it better!

robbieaverill:
Ok cool, thanks for confirming 👍

src/components/TextField.vue
```diff
@@ -84,6 +84,10 @@ export default {
     */
     multiline: Boolean,
     /**
+    * Force label to always be in the minimized position.
+    */
+    minimizedLabel: Boolean,
+    /**
     * Display text field in currency mode
     */
     currency: Boolean,
```
robbieaverill, line 93:
```suggestion
     * Force label to always be in the minimized position.
     */
    minimizedLabel: Boolean,
    /**
     * Display text field in currency mode
     */
    currency: Boolean,
```

robbieaverill (approved):

PR #312, opened 2020-09-30
Restore placeholders in text fields

### LINKED ISSUE #310: Stop using the placeholder attribute of form fields to render labels
We use the placeholder attribute in form fields to render labels, instead of using a label.

This means we can't actually have a placeholder, so fields like this can't be built at the moment without making the infinity symbol part of the field's value, which would be weird for the users because it'd be selectable (as well as being bound as part of the value).

image

We should change this.

ScopeyNZ:
Fixed #310

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -88,12 +88,17 @@ export default {
     */
     minimizedLabel: Boolean,
     /**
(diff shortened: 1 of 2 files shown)


robbieaverill (approved):
Ok

PR #313, opened 2020-10-01
Restore placeholders to dropdowns

ScopeyNZ:
#312 did this for text fields which was a bit of a functional requirement. This wasn't so pressing, but keeps feature parity between the two different fields.


kvisca (approved):
Pulled down and tested. LGTM!

PR #315, opened 2020-10-02
Update hologram background and add parallax directive

ScopeyNZ:
Side effect here is that we remove the unnecessary and huge `bg-hologram.png` from our bundle.


ScopeyNZ:
Hmmmn the accelerometer stuff is a bit janky.

ScopeyNZ:
FYI for the accelerometer stuff - it works, it's not buggy. "Janky" just means that sometimes the hologram effect jumps a bit if you orient your phone in a certain direction. I could probably tweak it some more, but I'm not very good at the maths involved.

jaepass:
This is sweet can't wait to see it implemented!

jaepass (approved):

jaepass:
I see what you mean with the slight jump on touch too but I don't think that's a huge issue. Looks great otherwise on desktop!

PR #317, opened 2020-10-04
Only load dropzone in browser context

ScopeyNZ:
Turns out dropzone is not SSR compatible. The recommended approach here (from a quick google) is to just conditionally load the package.


kvisca (approved):
LGTM

PR #320, opened 2020-10-07
Update tailwind and remove use of deprecated utilities

ScopeyNZ:


ScopeyNZ:
One thing I did _not_ experiment with yet is their new version of `@apply` which should work with `space-x` etc.

robbieaverill (approved):

PR #329, opened 2020-10-14
[CHEC-826] WYSIWYG Component

kvisca:
WYSIWYG component for the UI Library. Note, the package seems to be having an issue with Prosemirror as [documented here](https://github.com/ueberdosis/tiptap/issues/577).  I have added a workaround for the time being. 

We didn't have a style for the dark mode so I used my best judgement @blazepollard. Let me know if you'd like this adjusted.

Light Mode:
Screen Shot 2020-10-14 at 6 09 05 PM
Dark Mode:
Screen Shot 2020-10-14 at 6 09 12 PM


src/components/ChecWysiwyg.vue
```diff
@@ -0,0 +1,571 @@
+<template>
+  <div class="wysiwyg" :class="classNames">
```
robbieaverill, line 2:
```suggestion
  <div class="chec-wysiwyg" :class="classNames">
```

We've have conflicts with generic class names like this before

src/components/ChecWysiwyg.vue
```diff
@@ -0,0 +1,571 @@
...
+        v-if="required"
+        class="wysiwyg__required-text"
+      >
+        {{ $t('general.required') }}
+      </span>
+    </label>
+    <EditorMenuBar v-slot="{ commands, isActive }" :editor="editor">
+      <div class="wysiwyg__menubar">
+        <button
+          class="wysiwyg__menubar--button"
```
robbieaverill, line 26:
```suggestion
          type="button"
          title="Bold"
          class="wysiwyg__menubar--button"
```

Can you apply this to all of the buttons? BYO translations

src/components/ChecWysiwyg.vue
```diff
@@ -0,0 +1,571 @@
...
+          @click="commands.italic"
+        >
+          <ChecIcon icon="italic" size="base" />
+        </button>
+        <button
+          class="wysiwyg__menubar--button"
+          :class="{ 'wysiwyg__menubar--button--is-active': isActive.heading({ level: 1 }) }"
+          @click="commands.heading({ level: 1 })"
+        >
+          H1
```
robbieaverill, line 44: Any reason these aren't icons?

src/components/ChecWysiwyg.vue
```diff
@@ -0,0 +1,571 @@
...
+      editor: new Editor({
+        extensions: [
+          new Blockquote(),
+          new BulletList(),
+          new Heading({ levels: [1, 2, 3] }),
+          new Bold(),
+          new Italic(),
+          new ListItem(),
+        ],
+        content: '<p>This is just a boring paragraph</p>',
```
robbieaverill, line 175:
```suggestion
        content: '',
```

Seems like something that should be in the storybook

src/components/ChecWysiwyg.vue
```diff
@@ -0,0 +1,571 @@
...
+       */
+      this.autoGrow();
+      this.$emit('input', $event.target.value, $event);
+    },
+    adjustSlotWidth() {
+      this.hasSlot = this.$refs.rightContentSlot.clientWidth > 0;
+      this.slotWidth = this.$refs.rightContentSlot.scrollWidth;
+    },
+    autoGrow() {
+      if (!this.multiline) {
```
robbieaverill, line 277: Is this a copy paste from TextField? I assume WYSIWYG editors would always be multiline

src/components/ChecWysiwyg.vue
```diff
@@ -0,0 +1,571 @@
...
+        return;
+      }
+      /**
+       * Allows the text area to grow to match the value as the user is typing.
```
robbieaverill, line 281: Is there a max-height?

robbieaverill (commented):
There is a lot of copy and pasted code in the WYSIWYG component from TextField. Can you please centralise some of it, or use generic CSS class names so you can re-use those between components? Defining everything twice is not ideal

ScopeyNZ:
I played around with this a bit locally, and I found some issues.

I fixed the placeholders, and the value binding (so it emits change events). 

There are a few other issues that I've spotted:

- The auto-resize needs to kick in much earlier, as you end up typing under the button bar.
- The fade effect under the label that the textareas have isn't working, so when you scroll you can't read the label
- There's no blockquote style

src/components/ChecWysiwyg.vue
kvisca: They weren't icons in the design so I kept it that way.

src/components/ChecWysiwyg.vue
kvisca: Good call

ScopeyNZ:
TODO list:

[] Blockquote styles
[] Label underlay
[] Make the size configurable
[x] Fix the button bar overlap (remove `position: absolute`)
[] Placeholders

ScopeyNZ:
Ok. Squashed and rebased the whole thing and LGTM.

robbieaverill (approved):
Looks good and works well. Some notes:

* In dark mode the active buttons in the toolbar need to invert the icon/font colour, ATM it's dark on dark:
image
* I'd like it if you could resize the thing yourself in the browser, but that's not a blocker
* When you don't have a label, it still has space for the label. Not sure if that's intended or not, and it's not a blocker

ScopeyNZ:
Nice spotting. I'll raise an issue for those things as a bug, but I don't think any of them block us going ahead with the add-product screen. Thanks!

Issue #301, opened 2020-09-24
InnerBlock div requires flex-grow


kvisca:
InnerBlocks top-level div needs a flex-grow of 1 added to it so that the children display full width.
![Screen Shot 2020-09-24 at 3 42 24 PM](https://user-images.githubusercontent.com/36721153/94192211-a1ab9d80-fe7c-11ea-8a2b-a7c5cbe320a4.png)

`.extra-fields__block-extra > div {
    flex-grow: 1;
}`

ScopeyNZ:
Just for some context, this component has a structure like:

```
<div class="inner-card">
  <div>
    <slot />
  </div>
</div>
```

It does this because it _may_ have an action on the right hand side of the card. I think the first div within this component should have `flex-grow`.

robbieaverill:
Done in #332

PR #332, opened 2020-10-16
Adding flex grow to solve InnerBlock full width issue.

kvisca:
Adding flex-grow to solve issue causing repeated fields to not be full width if an action is on the right side.

Solves #301


ScopeyNZ:
Can we avoid inline tailwind classes and switch this to BEM?

ScopeyNZ (approved):
Nice.

robbieaverill:
@kvisca pro tip for future PRs: https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

Issue #328, opened 2020-10-14
Currency fields don't render correctly when currency symbol is longer than one character


robbieaverill:
Use Romanian RON for example:

![image](https://user-images.githubusercontent.com/5170590/96049993-c23c9700-0e2d-11eb-9a01-b4e42b1b8700.png)

robbieaverill:
It actually makes it the product dashboard pretty unusable:

![image](https://user-images.githubusercontent.com/5170590/96050915-578c5b00-0e2f-11eb-8bfc-aabbf5788315.png)

robbieaverill:
cc @ScopeyNZ @blazepollard

robbieaverill:
Done in #333

PR #333, opened 2020-10-16
Making input padding dynamic based on currency

kvisca:
Making the input padding dynamic for currencies. 

Screen Shot 2020-10-16 at 1 31 10 PM
Screen Shot 2020-10-16 at 1 31 04 PM
Screen Shot 2020-10-16 at 1 30 57 PM

Solves #328

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -34,7 +34,7 @@
         {{ $t('general.required') }}
       </span>
     </label>
-    <div class="text-field__currency">
+    <div ref="currency" class="text-field__currency">
       {{ currencySymbol }}
     </div>
     <div v-if="$slots.default" ref="rightContentSlot" class="text-field__right-content">
@@ -194,6 +194,11 @@ export default {
         this.autoGrow();
       });
     },
+    currencySymbol() {
+      this.$nextTick(() => {
+        this.adjustCurrencyPadding();
+      });
+    },
     isScrollable(scrollable) {
       if (scrollable) {
         this.$refs.input.addEventListener('scroll', this.handleScroll);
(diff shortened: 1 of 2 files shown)


src/components/TextField.vue
```diff
@@ -235,6 +240,12 @@ export default {
       this.hasSlot = this.$refs.rightContentSlot.clientWidth > 0;
       this.slotWidth = this.$refs.rightContentSlot.scrollWidth;
     },
+    adjustCurrencyPadding() {
```
robbieaverill, line 243: Can you add a doc block to explain what this method does pleasee?

src/components/TextField.vue
```diff
@@ -235,6 +240,15 @@ export default {
...
+      if (!this.currency) {
+        return;
+      }
+      /**
+       * Allows the input padding to grow to match the width of the currency symbol container.
+       */
+      this.$refs.input.style.paddingLeft = `${this.$refs.currency.offsetWidth}px`;
+    },
```
robbieaverill, line 251:
```suggestion
    /**
     * Allows the input padding to grow to match the width of the currency symbol container.
     */
    adjustCurrencyPadding() {
      if (!this.currency) {
        return;
      }
      this.$refs.input.style.paddingLeft = `${this.$refs.currency.offsetWidth}px`;
    },
```

ScopeyNZ:
I was going to ask why we're doing this with an absolutely positioned div rather than just putting it next to the input, but I see that the input is actually the full size, which makes sense from a click to focus point of view.

We could probably change this in the future to remove some of this JS but LGTM for now.

ScopeyNZ (approved):

Issue #318, opened 2020-10-05
Currency text fields with zero value are missing the currency label


ScopeyNZ:
![image](https://user-images.githubusercontent.com/3260989/95140605-22438700-0724-11eb-8c10-731513ae98d0.png)

There's some logic which is casting the value to a bool, which isn't correct. It appears that this logic is driving the `--modified` modifier, which seems opposite to the existing `--empty` modifier that has a better name and has different logic. Maybe we can remove the `--modified` modifier and just use `--empty`?

robbieaverill:
Done in #334

PR #334, opened 2020-10-16
Allowing zero to show currency field

kvisca:
Zero will now show the currency icon if the field is a currency type.

Screen Shot 2020-10-16 at 4 40 18 PM

Solves #318

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -172,7 +172,7 @@ export default {
         minimizedLabel,
       } = this;
 
-      return {
+      const classes = {
         'text-field--currency': currency,
         'text-field--disabled': variant === 'disabled',
         'text-field--error': variant === 'error',
@@ -183,6 +183,15 @@ export default {
         'text-field--minimized-label': minimizedLabel,
         [`text-field--${styleVariant}`]: styleVariant !== '',
       };
+
+      /**
+       * If the input is a currency field, allow 0 to show in the modified state.
+       */
+      if (currency && value === 0) {
+        classes['text-field--modified'] = true;
+      }
+
+      return classes;
     },
     hasPlaceholder() {
       return typeof this.placeholder === 'string' && this.placeholder.length > 0;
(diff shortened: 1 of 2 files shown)


src/components/TextField.vue
```diff
@@ -183,6 +183,15 @@ export default {
         'text-field--minimized-label': minimizedLabel,
         [`text-field--${styleVariant}`]: styleVariant !== '',
       };
+
+      /**
+       * If the input is a currency field, allow 0 to show in the modified state.
+       */
+      if (currency && value === 0) {
+        classes['text-field--modified'] = true;
+      }
```
robbieaverill, line 192: I don't understand this logic - why do you need to mark the field as modified if it has a value of 0?

kvisca: Because technically it is modified in this case. The value is just zero. It needs to be marked as Modified so that the Currency symbol is visible as well as any other styling that's inherited by an input being marked as such.  Hopefully, that makes sense haha

robbieaverill: What I'm reading is that if I render `<TextField value="0" currency />` in a template, it will have the `text-field--modified` class added to it before I've modified anything - that feels odd

kvisca: Hmm okay, I can see how that feels odd. I was looking at it not from a user modifying it but the input being modified (in this case by the value being present). I'll see if I can break it out to make it a bit cleaner.

ScopeyNZ: I think we should drop "modified" as we already have "empty". It seems like "modified" is a badly named opposite to "empty".

ScopeyNZ:
I'm not sure if you noticed my comment. How is this not just the opposite of `--empty`?

kvisca:
> I'm not sure if you noticed my comment. How is this not just the opposite of `--empty`?

Completely. Let me know if this is better.

ScopeyNZ (approved):
👌

PR #340, opened 2020-10-20
Add "text" (or "label") button style/variant

ScopeyNZ:
Required for the new design for the delete button in add product

### DIFF
--- a/src/components/ChecButton.vue
+++ b/src/components/ChecButton.vue
@@ -71,7 +71,7 @@ export default {
     /**
      * Changes the button to a transparent button with coloured text that matches the chosen colour.
      */
-    text: Boolean,
+    textOnly: Boolean,
     /**
      * Disables the button
      */
@@ -92,7 +92,7 @@ export default {
         `button--color-${this.color}`,
         `button--variant-${this.variant}`,
         {
-          'button--text': this.text,
+          'button--text-only': this.textOnly,
           'button--disabled': this.disabled,
           'button--has-icon': this.hasIcon,
           [`button--has-icon-${this.iconPosition}`]: this.hasIcon,
@@ -289,7 +289,7 @@ export default {
           @apply bg-#{map.get($config, 'active')};
         }
 
-        &.button--text {
+        &.button--text-only {
           // Handle the one gradient that can't be text
           $default-color: #{map.get($config, 'default')};
           @if ($default-color == 'primary-gradient') {
@@ -314,7 +314,7 @@ export default {
     }
   }
 
-  &--text {
+  &--text-only {
     @apply border-0 shadow-none;
 
     &:focus {
--- a/src/components/ChecButton.vue
+++ b/src/components/ChecButton.vue
@@ -39,18 +39,21 @@ export default {
           'blue',
           'green',
           'red',
+          'purple',
+          'orange',
+          'white',
+          'gray',
         ].includes(color);
       },
       default: 'secondary',
     },
     /**
-     * The style variant of the button. Includes size, border radius, etc. One of 'large', 'regular', 'small',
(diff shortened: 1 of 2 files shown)


robbieaverill:
Linting failure

src/components/ChecButton.vue
```diff
@@ -39,18 +39,21 @@ export default {
...
     /**
-     * The style variant of the button. Includes size, border radius, etc. One of 'large', 'regular', 'small',
-     * 'round', 'tag'
+     * The style variant of the button. Includes size, border radius, etc. One of 'regular', 'small', 'round', 'tag'
      */
     variant: {
       type: String,
       validate(variant) {
-        return ['large', 'regular', 'small', 'round', 'tag'].includes(variant);
+        return ['regular', 'small', 'round', 'text', 'tag'].includes(variant);
```
robbieaverill, line 56: Did you mean to remove large?

Is it used anywhere in the dashboard or auth service?

ScopeyNZ: Oh yeah, I forgot to add that to the description. It's not used, and has been removed from Figma.

ScopeyNZ:
Any further feedback?

robbieaverill:
Any reason you didn't make this a variant?

The "brand" colour doesn't look right with the text colour:
image

kvisca:
Seems weird to me to be a prop called text. If I saw that in the mark up I would think it's the actual text of the button. Maybe rename it to `textVariant` or `textOnly`.  

Otherwise, LGTM minus the brand variant not having a text-only version.

ScopeyNZ:
I didn't make it a variant as I figured it has to match the size of other buttons with different variants. The "variant" as-is should probably be called "size".

Fixed the brand colour issue.

I didn't really like the naming either - but I really struggled to come up with something better. I considered `text-variant` but it felt weird with the `variant` prop too.

I could do a prop called `style` with the options being `text` and `button` but that also felt like it conflicts with `variant`.

ScopeyNZ:
I've updated it to `textOnly`.

robbieaverill:
You're essentially replicating the ["link" button style](https://getbootstrap.com/docs/4.4/components/buttons/) from Bootstrap. Their naming works because it looks exactly like a link when it renders. I agree with Kyle that text (and textOnly) are not good names for this prop.

What about "outline" or something?

kvisca:
> I didn't make it a variant as I figured it has to match the size of other buttons with different variants. The "variant" as-is should probably be called "size".
> 
> Fixed the brand colour issue.
> 
> I didn't really like the naming either - but I really struggled to come up with something better. I considered `text-variant` but it felt weird with the `variant` prop too.
> 
> I could do a prop called `style` with the options being `text` and `button` but that also felt like it conflicts with `variant`.

I think `variant` should be renamed `style` and the `textOnly` to `variant` with a type of text or button like you suggested. That makes a lot more sense.

robbieaverill:
I agree with Kyle, but not "style" because that's a native HTML attribute name

ScopeyNZ:
I agree that variant should be renamed, probably to `size` or something, but that's a lot more effort 😢 . I'd have to PR all the repos that use buttons to change everything. Do you think I can raise an issue to do that outside of this PR?

Happy to rename the existing prop to whatever makes the most sense until then though.

kvisca:
> I agree that variant should be renamed, probably to `size` or something, but that's a lot more effort 😢 . I'd have to PR all the repos that use buttons to change everything. Do you think I can raise an issue to do that outside of this PR?
> 
> Happy to rename the existing prop to whatever makes the most sense until then though.

Yea I think as long as we have an issue to rename it we're good. A couple of components have this odd variant naming issue, so we could attack them all at once then.

PR #352, opened 2020-10-27
[CHEC-977] Updating tag button style, removing light grey tag type

### LINKED ISSUE #351: Update ChecTagGroup button style
Looks like this:
image

New designs look like this:
image

kvisca:
- Updated the tag button to the square variant now used for tag groups
- Removed the light grey tag type as per discussion with Blaze (It will default to the whit type if used).

Screen Shot 2020-10-27 at 1 43 03 PM

Resolves #351

### DIFF
--- a/src/components/ChecButton.vue
+++ b/src/components/ChecButton.vue
@@ -228,10 +228,6 @@ export default {
     &-tag {
       @apply py-2 px-4 rounded leading-tight;
 
-      // Tag button padding override
-      padding-bottom: 0.625rem;
-      padding-top: 0.625rem;
-
       .button__content {
         @apply caps-xxs;
       }
--- a/src/components/ChecButton.vue
+++ b/src/components/ChecButton.vue
@@ -228,6 +228,10 @@ export default {
     &-tag {
       @apply py-2 px-4 rounded leading-tight;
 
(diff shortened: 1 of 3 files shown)


robbieaverill:
The button in your screenshot isn't the same height as the tags

kvisca:
> The button in your screenshot isn't the same height as the tags

As per Figma, it's not supposed to be. I checked with Blaze as well.

Screen Shot 2020-10-27 at 1 55 04 PM

robbieaverill:
Cool!

robbieaverill:
Blocking feedback:

* Font size is 15px, should be 12px (see https://github.com/chec/ui-library/issues/351)
* In Figma the bottom of the "X" button is a couple of pixels lower than the bottom of the tag text, and the top is a couple of pixels above the top of the tag text - both need to be updated here (could be fixed by changing the font size)
* The white tag has a 1px grey border which isn't in the designs
* The disabled state is 1px smaller (missing border?) than the non-disabled state which causes the UI to jump slightly

Non-blocking feedback:

* The storybook knob "color" is dark-grey by default, but renders as white

robbieaverill (approved):
LGTM, nice!

PR #360, opened 2020-11-05
[CHEC-968] Code block groups

kvisca:
- Updating to new package to allow for line numbers.
- Introducing code block group component. 
Screen Shot 2020-11-04 at 7 46 14 PM

### DIFF
--- a/src/components/CodeBlock.vue
+++ b/src/components/CodeBlock.vue
@@ -1,35 +1,98 @@
 <template>
-  <Prism
+  <PrismEditor
+    v-model="code"
     class="code-example"
-    :class="{ 'language-none': language === 'none' ? 'language-none' : '' }"
-    :language="language !== 'none' ? language : 'javascript'"
+    :class="classNames"
+    :highlight="highlighter"
+    :line-numbers="lineNumbers"
+    :readonly="!editable"
   >
+    <!--
+      @slot Codeblock content
+    -->
     <slot />
-  </Prism>
+  </PrismEditor>
 </template>
 
 <script>
-// Base Prism import
-import 'prismjs';
-// Extra language configuration
-import 'prismjs/components/prism-markup-templating';
-import 'prismjs/components/prism-jsx';
-import 'prismjs/components/prism-json';
-import 'prismjs/components/prism-php';
 // Vue component for Prism
-import Prism from 'vue-prism-component';
+import { PrismEditor } from 'vue-prism-editor';
+import 'vue-prism-editor/dist/prismeditor.min.css';
+// Prism core
+import { highlight, languages } from 'prismjs/components/prism-core';
+// Used as a basis for langauge packs.
+import 'prismjs/components/prism-clike';
+// Import supported languages
+import 'prismjs/components/prism-markup-templating';
+import 'prismjs/components/prism-javascript';
+import 'prismjs/components/prism-json';
+import 'prismjs/components/prism-markup';
+import 'prismjs/components/prism-jsx';
+import 'prismjs/components/prism-php';
 
 import 'prismjs/themes/prism.css';
 
 export default {
   name: 'CodeComponent',
   components: {
-    Prism,
+    PrismEditor,
   },
   props: {
+    /**
+     * The language for syntax highlighting.
+     *  Accepts either none, javascript, json, jsx or php.
+     */
     language: {
       type: String,
+      validate(type) {
+        return ['none', 'javascript', 'json', 'jsx', 'php'].includes(type);
+      },
       default: 'none',
+    },
+    /**
+     * Display codeblock line numbers.
+     */
+    lineNumbers: Boolean,
+    /**
+     * Allow the codeblock to grow or shrink with the content
+     */
+    grow: Boolean,
+    /**
+     * Toggles on the editable mode.
+     */
+    editable: Boolean,
+  },
+  data() {
+    return {
+      code: '',
+    };
+  },
+  computed: {
+    classNames() {
+      const {
+        language,
+        lineNumbers,
+      } = this;
+
+      const classes = {
+        'code-example--line-numbered': lineNumbers,
+        'code-example--fixed': !this.grow,
+        'language-none': language === 'none',
+      };
+
+      return classes;
+    },
+  },
+  mounted() {
+    // On mount, add the slot data to the data object.
+    this.code = this.$slots.default[0].text;
+  },
+  methods: {
+    highlighter(code) {
+      if (this.language === 'none') {
+        return highlight(code, languages.markup);
+      }
+      return highlight(code, languages[this.language]);
     },
   },
 };
@@ -44,7 +107,7 @@ pre.code-example[class*='language-'] {
 
 // Theme
 .code-example {
-  @apply bg-code-block rounded-md px-5 overflow-auto font-mono bg-gray-600 text-gray-100 w-full;
+  @apply bg-code-block rounded-md px-5 py-6 overflow-auto font-mono bg-gray-600 text-gray-100 w-full;
 
   pre {
     margin: 0.5em 0;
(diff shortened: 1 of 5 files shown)


src/components/CodeBlock.vue
```diff
@@ -1,36 +1,90 @@
...
 export default {
   name: 'CodeComponent',
   components: {
-    Prism,
+    PrismEditor,
   },
   props: {
+    /**
+     * The language for syntax highlighting.
+     *  Accepts either none, javascript, json, jsx or php.
```
robbieaverill, line 42: You could make this a validate() function if you wanted to enforce it

src/stories/components/CodeBlock.stories.mdx
```diff
@@ -21,10 +24,75 @@ import CodeBlock from '../../components/CodeBlock.vue';
...
+  </Story>
+</Preview>
+
+## Code Block Group
+### Required attributes to properly render group:
+`Title` The tabs in the tab group will be generated based on the titles passed to each codeblock.
+
+`Key` The key is required for proper rerendering of components when switching active blocks.
+
+### Example implimentation:
```
robbieaverill, line 51:
```suggestion
### Example implementation
```

src/stories/components/CodeBlock.stories.mdx
```diff
@@ -21,10 +24,75 @@ import CodeBlock from '../../components/CodeBlock.vue';
...
+      template: `
+        <div class="p-16 flex justify-center w-1/2 m-auto" :class="{'bg-gray-500': background === 'dark'}">
+          <CodeBlock :language="language" :editable="editable" :line-numbers="lineNumbers">{{ code }}</CodeBlock>
+        </div>`
+    }}
+  </Story>
+</Preview>
+
+## Code Block Group
+### Required attributes to properly render group:
```
robbieaverill, line 46:
```suggestion

### Required attributes

```

src/stories/components/CodeBlock.stories.mdx
```diff
@@ -21,10 +24,75 @@ import CodeBlock from '../../components/CodeBlock.vue';
...
+        editable: {
+          default: boolean('Editable Block', false),
+        },
+        background: {
+          default: select('Background Color', ['light', 'dark']),
+        },
+      },
+      template: `
+        <div class="p-16 flex justify-center w-1/2 m-auto" :class="{'bg-gray-500': background === 'dark'}">
+          <CodeBlock :language="language" :editable="editable" :line-numbers="lineNumbers">{{ code }}</CodeBlock>
```
robbieaverill, line 39: Changing this in the storybook doesn't update the preview

src/stories/components/CodeBlock.stories.mdx
```diff
@@ -21,10 +24,75 @@ import CodeBlock from '../../components/CodeBlock.vue';
...
+      },
+      props: {
+        language: {
+          default: select('Language', ['none', 'javascript', 'json', 'jsx', 'php']),
+        },
+        code: {
+          default: text('Code', "// Render your payment form\n\nconst elements = stripe.elements();\n\nconst card = elements.create(1234);"),
+        },
+        lineNumbers: {
+          default: boolean('Line Numbers', true),
```
robbieaverill, line 80: This doesn't seem to change anything in the storybook

Also:
```suggestion
          default: boolean('Line numbers', true),
```

robbieaverill (commented):
Nice, looking pretty good! 

The designs suggest that the code block should scroll rather than wrap. There should also be no right or bottom padding (see https://www.figma.com/file/QapWtySLg7GOXEdGcTFVDZ?node-id=6579:780#46675787).

kvisca:
> Nice, looking pretty good!
> 
> The designs suggest that the code block should scroll rather than wrap. There should also be no right or bottom padding (see https://www.figma.com/file/QapWtySLg7GOXEdGcTFVDZ?node-id=6579:780#46675787).

Made the updates requested. Spoke with @blazepollard and we defined min and max heights with a custom scrollbar. Also now has an override in the event we want it to grow with the content.

src/stories/components/CodeBlock.stories.mdx
kvisca: I can only reproduce this on the code block group and I believe that's because of how the vnode is being rendered. You will need to switch tabs to see the updates. Would we ever be dynamically changing these values? I'm having a hard time finding a solution.

robbieaverill: Na I doubt it. Maybe just remove the bits that don't bind dynamically from the story

robbieaverill:
@kvisca there's a build error

src/components/CodeBlock.vue
```diff
@@ -1,36 +1,98 @@
...
-// Extra language configuration
+// Vue component for Prism
+import { PrismEditor } from 'vue-prism-editor';
+import 'vue-prism-editor/dist/prismeditor.min.css';
+// Prism core
+import { highlight, languages } from 'prismjs/components/prism-core';
+// Import supported languages
 import 'prismjs/components/prism-markup-templating';
-import 'prismjs/components/prism-jsx';
+import 'prismjs/components/prism-clike';
```
robbieaverill, line 25: It feels like we'll probably not need this language. I assume that not including it here will reduce our bundle size slightly, maybe remove it?

src/components/CodeBlock.vue
```diff
@@ -54,6 +116,27 @@ pre.code-example[class*='language-'] {
     white-space: pre-wrap; // override prism.js tomorrow-theme default
   }
 
+  &--line-numbered {
+    @apply pl-0;
+  }
+
+  &--fixed {
+    min-height: 20rem;
+    max-height: 35rem;
```
robbieaverill, line 125: Where do these numbers come from? Are they related to the dashboard's content container width? If we can't avoid having these here, it'd be good to comment what they reference/why they're here.

src/components/CodeBlockGroup.vue
```diff
@@ -0,0 +1,68 @@
...
+    };
+  },
+  computed: {
+    tabs() {
+      // Creates a list of titles based on the slot vnodes
+      const groups = this.$slots.default.filter((slot) => typeof slot.data !== 'undefined')
+        .map(slot => slot.data.attrs.title);
+      return groups;
+    },
+    // Return the vnode based on the selected tab.
```
robbieaverill, line 43:
```suggestion
    /**
     * Return the vnode based on the selected tab.
     */
```

robbieaverill (commented):
Nice! The storybook still has inputs for the code, which can't be changed (remove them) and has an "Allow Grow" knob on the main story, which feels like it should be a boolean field - it's currently a dropdown with no options in it

src/components/CodeBlock.vue
kvisca: These are just the numbers Blaze and I came up with. I will add a comment.

src/components/CodeBlock.vue
kvisca: the clike component isn't a language, it's the basis used by other language packs like PHP.  I've moved it to the top to help denote that as well as left a comment.

robbieaverill: Okay, it needs to be there though? We haven't needed it in the past

kvisca: This package seems to need it otherwise it breaks

ScopeyNZ:
Did you figure out why the code knob isn't working? I don't think we have a use-case for it right now, but it might be an issue in the future if we have dynamic content.

src/components/CodeBlock.vue
```diff
@@ -1,36 +1,99 @@
...
-    :language="language !== 'none' ? language : 'javascript'"
+    :class="classNames"
+    :highlight="highlighter"
+    :line-numbers="lineNumbers"
+    :readonly="!editable"
   >
+    <!--
+      @slot Codeblock content
+    -->
     <slot />
```
robbieaverill, line 13: Just out of interest, does this mean you're binding the `code` data property to the PrismEditor instance with `v-model` while also passing it as a slot?

kvisca: It's to allow for us to still use the slot like the previous integrations.

kvisca:
> Did you figure out why the code knob isn't working? I don't think we have a use-case for it right now, but it might be an issue in the future if we have dynamic content.

I think we'd have to stop using slots then. I did it this way to avoid reintegrating since we don't actually allow inline editing.

ScopeyNZ:
Right okay. Let's just disable the knob for now then.

kvisca:
> Right okay. Let's just remove the knob for now then.

Removed.

src/components/CodeBlock.vue
ScopeyNZ: I think this is a good use-case for a functional component. You're just returning a PrismEditor component, and you don't actually need to put the slot in the template - you're just accessing it's content with code. It might address the issues you were having with state updating too.

ScopeyNZ (approved):
I'm happy. But I have this feeling that there's some weirdness in this component I can't really shake. That `key` issue you were having included. I'd be interested if the same problems happen if this is a functional component.

I think we can merge right now though, as it works well enough - if we have problems with this component in the future though we have some ideas to go with.

PR #365, opened 2020-11-16
Minor popover bugfixes

ScopeyNZ:
- Ensure correct DOM node is used for popper.js
- Add menu name to MountingPortal
- Ensure popper target ref exists before use with poppoer.js

Part of fixing the cypress tests for chec/dashboard#463 and fixing the ui-library for use in the dash.


robbieaverill (approved):
LGTM

PR #367, opened 2020-11-26
Removing event prop on router links if button is disabled

kvisca:
If a button is set as a router link and also disabled, the button will still function. Natively, Vue router does not support disabling router links. A work around to this is to set the event from "click" to "" to achieve the desired effect.

### DIFF
--- a/src/components/ChecButton.vue
+++ b/src/components/ChecButton.vue
@@ -104,9 +104,15 @@ export default {
     },
     tag() {
       if (this.tagType === 'link') {
+        if (this.disabled) {
+          return 'span';
+        }
         return 'a';
       }
       if (this.tagType === 'route') {
+        if (this.disabled) {
+          return 'span';
+        }
         return 'router-link';
       }
       return 'button';


robbieaverill (approved):
Some context would be nice. I'm approving blindly because I trust you.

kvisca:
> Some context would be nice. I'm approving blindly because I trust you.

You're right, I updated the description to flush it out. Thanks @robbieaverill

robbieaverill:
> If a button is set as a router link and also disabled, the button will still function

I have a feeling that the right way to fix that problem is to render a span instead of an anchor, since anchors don't have a disabled attribute. Removing the event feels like a bit of a patchy solution to that. Semantically you still have an anchor with an href that the browser/screenreaders/etc will interpret as a valid link.

kvisca:
> > If a button is set as a router link and also disabled, the button will still function
> 
> I have a feeling that the right way to fix that problem is to render a span instead of an anchor, since anchors don't have a disabled attribute. Removing the event feels like a bit of a patchy solution to that. Semantically you still have an anchor with an href that the browser/screenreaders/etc will interpret as a valid link.

I can agree with that approach, I'll make the necessary changes.

src/components/ChecButton.vue
```diff
@@ -146,6 +152,9 @@ export default {
     const props = {};
     if (this.tagType === 'route') {
       props.to = this.$attrs.to;
+      if (this.disabled) {
+        props.event = '';
+      }
```
robbieaverill, line 157: Don't need this any more right?

robbieaverill (dismissed):
Remove the initial code for this and leave the span change

robbieaverill:
I wonder if the deploy preview will show up at some stage...

robbieaverill (approved):
The hover effect is still active when the "link" type button is disabled, but this isn't a regression in this PR

PR #371, opened 2020-12-08
Image manager adding max file limit and alternative column structure

kvisca:
- Adding user-defined maximum file limit.
- Adding a 2 column layout needed in the sidebar
- Adding responsive for 4 column layout.

**2 Column:**
Screen Shot 2020-12-08 at 5 35 43 PM

**3 file max:**
Screen Shot 2020-12-08 at 5 36 15 PM

### DIFF
--- a/src/components/ChecImageManager.vue
+++ b/src/components/ChecImageManager.vue
@@ -4,7 +4,7 @@
       <Draggable
         v-if="allFiles.length"
         class="image-rows-container"
-        :class="{ 'image-rows-container--dragging': dragging }"
+        :class="classNames"
         :value="allFiles"
         filter=".chec-image-item--loading"
         @input="reorder"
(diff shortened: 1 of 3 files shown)


robbieaverill:
What happens with max = 1?

kvisca:
> What happens with max = 1?

You'll be able to have 1 file, which is what we want for the editor. 

Screen Shot 2020-12-08 at 5 46 37 PM

robbieaverill:
Yum

robbieaverill (commented):
Is there a difference between columns and maxFiles? Maybe maxFiles needs to be defined as a prop?

kvisca:
> Is there a difference between columns and maxFiles? Maybe maxFiles needs to be defined as a prop?

Columns you can have either 2 or 4 column layouts. 

`maxFiles` is a prop that's completely separate from the columns. I didn't add it as a knob because it was acting very oddly if dynamically changed, assuming that's what you're referring to?

robbieaverill:
No, it needs to be defined as a prop on the component that uses it

src/components/ChecImageManager.vue
```diff
@@ -24,11 +24,15 @@
           @remove="() => removeFile(file)"
         />
       </Draggable>
-      <ChecButton @click="openDialog">
-        <ChecIcon class="chec-image-manager__icon" icon="image" size="base" /> {{ $t('imageManager.chooseImages') }}
-      </ChecButton>
-      <div v-if="footnote" class="chec-image-manager__helper">
-        {{ footnote }}
+      <div v-if="maxFiles === null || maxFiles > allFiles.length">
```
ScopeyNZ, line 27:
```suggestion
      <template v-if="maxFiles === null || maxFiles > allFiles.length">
```

BYO closing tag. Makes more sense not to add the block if it isn't needed?

src/components/ChecImageManager.vue
```diff
@@ -52,16 +56,52 @@ export default {
...
       default: null,
     },
+    /**
+     * Number of columns to be displayed. either 2 or 4.
+     */
+    columns: {
+      type: Number,
+      default: 4,
+      validator(value) {
+        return [2, 4].includes(value);
```
ScopeyNZ, line 73: Did you need this validator? Seems like an unnecessary restriction...

robbieaverill (approved):

src/components/ChecImageManager.vue
kvisca: Good call, I've switched it out.

src/components/ChecImageManager.vue
kvisca: No, I suppose not, I've removed it.

PR #373, opened 2020-12-10
Centring value if no label preset in text field

kvisca:
If no label is present, the value will be centered in the input. 
Screen Shot 2020-12-10 at 3 57 48 PM

### DIFF
--- a/src/components/TextField.vue
+++ b/src/components/TextField.vue
@@ -180,6 +180,7 @@ export default {
         'text-field--error': variant === 'error',
         'text-field--empty': value === '',
         'text-field--inline-label': label,
+        'text-field--labelless': label.length === 0,
         'text-field--multiline': multiline,
         'text-field--minimized-label': minimizedLabel,
         [`text-field--${styleVariant}`]: styleVariant !== '',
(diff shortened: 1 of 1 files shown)


src/components/TextField.vue
```diff
@@ -459,6 +460,12 @@ export default {
         @apply opacity-50 transition-opacity duration-150;
       }
     }
+
+    &.text-field--labelless {
+      .text-field__input {
+        @apply py-4;
+      }
+    }
```
robbieaverill, line 468: Can you move these definitions out of the scope they're in to reduce the level of nesting? In theory this should be enough on its own:

```scss
.text-field--labelless {
  .text-field__input {
    @apply py-4;
  }
}
```

robbieaverill (approved):
LGTM, minor suggestion/question

src/components/TextField.vue
kvisca: I tried that before but sadly that won't work without me setting those fields to important. I think this is cleaner for now.

robbieaverill: Yeah well that problem is only going to get worse if we keep doing this. Fine for now though.

kvisca: I'll open an issue to refactor the CSS for that c so we don't forget.

Issue #372, opened 2020-12-10
Update code block and group designs: buttons and static logos


robbieaverill:
![image](https://user-images.githubusercontent.com/5170590/101706932-0bc8ed00-3a3f-11eb-8b83-6e90e89e49e3.png)

Changes required:

* Change the buttons to be rounded
* Ensure the container around the buttons is also rounded and has the designed margin
* Add the ability for the code blocks to have static logos in the bottom right corner
  * For now just a single prop to say yes/no whether to show a static list of these
  * The logos should be static, and the content in the code blocks underneath should scroll underneath the logos (logos = absolute positioned)
* On small viewports, make the text in the buttons slightly smaller, and reduce the padding slightly, so it can handle three buttons in a row (note: this might be slightly different when the buttons are rounded)
![image](https://user-images.githubusercontent.com/5170590/101707090-5fd3d180-3a3f-11eb-950f-2b72ddbc637a.png)

When complete, pull in and finish https://github.com/chec/marketing-chec/issues/419

PR #375, opened 2020-12-11
Update code block tab designs and add option to display static framework logos in group

robbieaverill:
Resolves https://github.com/chec/ui-library/issues/372


src/assets/svgs/frameworks/angular.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="32" height="32" viewBox="0 0 32 32" fill="none" xmlns="http://www.w3.org/2000/svg">
+    <path d="M13.3761 16.864H16.0961H18.8161L16.0961 10.32L13.3761 16.864Z" fill="#D6E4F7"/>
```
kvisca, line 2: Fill should be ` fill="currentColor"` so we can control the color more easily. This applies to all the SVGs added

kvisca (changes requested):
Besides that one change, LGTM!

src/assets/svgs/frameworks/angular.svg
robbieaverill: Where do we define `currentColor`?

robbieaverill: It's a convention thing, I see. The Vue logo has two colours in it, how do you suggest I handle that? 
image

kvisca: The fill would be on the <path> (end of line 2 in that example).  The two-tone one is a bit of a challenge. I suppose we could do a `fill="currentcolor"` and  a `fill-opacity="0.4"` and try to match it as close as we can.

robbieaverill: How would you feel about me making the second line currentColor and leaving the other one the way it is? There'll be a phase 2 of this PR at some point in the future to make the logos configurable

robbieaverill: Actually your suggestion works perfectly, I'm going to use that

kvisca (approved):
Awesome LGTM!

PR #377, opened 2020-12-11
Updating accordion styles

kvisca:
Updating the accordion styling to match updated designs. 

Screen Shot 2020-12-11 at 1 54 55 PM

### DIFF
--- a/src/components/ChecAccordion.vue
+++ b/src/components/ChecAccordion.vue
@@ -144,11 +144,10 @@ export default {
 
 <style lang="scss">
 .accordion {
-  @apply rounded bg-gray-100;
+  @apply bg-white rounded-md shadow-sm mb-2;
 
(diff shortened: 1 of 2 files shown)


robbieaverill (approved):
Is that story you deleted no longer relevant?

kvisca:
> Is that story you deleted no longer relevant?

Correct. I did the changes with Blaze and moving forward we won't be using that design pattern.

ScopeyNZ:
Just chiming in here - that story may not have been a real world use case, but was highlighting issues with margin collapsing that came up at one point. The story was probably still relevant to illustrate that issue that might occur.

kvisca:
> Just chiming in here - that story may not have been a real world use case, but was highlighting issues with margin collapsing that came up at one point. The story was probably still relevant to illustrate that issue that might occur.

I'll put in a PR to add it back in then.

PR #379, opened 2020-12-12
[CHEC-1096] Adding round variant and dark styling for checkboxes

kvisca:
- Adds rounded variant of checkbox
- Adds dark variation of both default and rounded checkboxes.

**Rounded Variant:**
Screen Shot 2020-12-11 at 9 18 23 PM
**Rounded Variant Dark:** 
Screen Shot 2020-12-11 at 9 18 17 PM
**Default Dark styling:**
Screen Shot 2020-12-11 at 9 18 09 PM

### DIFF
--- a/src/components/ChecCheckbox.vue
+++ b/src/components/ChecCheckbox.vue
@@ -69,6 +69,13 @@ export default {
       default: '',
     },
     /**
+     * Variant style
+     */
+    variant: {
+      type: String,
+      default: 'default',
+    },
+    /**
      * Puts checkbox into an indeterminate state
      */
     indeterminate: Boolean,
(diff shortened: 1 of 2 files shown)


robbieaverill:
It almost feels to me as if the round ticks in these designs are icons rather than checkboxes, to be honest

ScopeyNZ:
It feels inconsistent to me to have different font for radio's versus checkboxes

kvisca:
> It almost feels to me as if the round ticks in these designs are icons rather than checkboxes, to be honest

I don't think I agree here since they specifically refer to the selection of a theme. They are in essence fancy radio boxes.

> It feels inconsistent to me to have different font for radio's versus checkboxes

I can agree with this, @blazepollard what do you think about normalizing the font between the two styles?

robbieaverill:
Ok, should the changes be made to ChecRadio instead then?

My reason for saying they could be icons is because the card containing these things feels like the UI element that needs to be created. The radio button with a tick in it is not a design pattern we use anywhere else yet. If it's going to become one then we can continue with doing it this way.

blazepollard:
I don't see us using that checkbox standalone. It would be attached to a card only. I see this as a variant of the card almost (selector card) or something. 

@kvisca @ScopeyNZ

blazepollard:
This box wouldn't be in a traditional form in other words.

robbieaverill:
Ok, that seems like a good justification to make this a dashboard only change then (adding the card element and using icons to control the checkbox effect). What do you think @kvisca?

kvisca:
> Ok, that seems like a good justification to make this a dashboard only change then (adding the card element and using icons to control the checkbox effect). What do you think @kvisca?

Okay, I will close this and add it to the dashboard.

PR #384, opened 2020-12-15
Adding outline button option

kvisca:
Adding outlined button option needed for marketing site.

**Light BG:**
Screen Shot 2020-12-15 at 1 00 40 PM

**Dark BG:**
Screen Shot 2020-12-15 at 1 02 25 PM

### DIFF
--- a/src/components/ChecButton.vue
+++ b/src/components/ChecButton.vue
@@ -73,6 +73,10 @@ export default {
      */
     textOnly: Boolean,
     /**
+     * Changes the button to an outline only button.
+     */
+    outline: Boolean,
+    /**
      * Disables the button
      */
     disabled: Boolean,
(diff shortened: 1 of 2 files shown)


robbieaverill:
The third line of the light BG screenshot is unreadable, as is the second line in the dark BG example. The first, and second to last (at least) in the dark BG won't pass contrast standards. Are we sure this is the correct implementation?

kvisca:
> The third line of the light BG screenshot is unreadable, as is the second line in the dark BG example. The first, and second to last (at least) in the dark BG won't pass contrast standards. Are we sure this is the correct implementation?

@robbieaverill you wouldn't use those variations on those backgrounds. Problem is, we need a white-bordered button outline but that wouldn't work on white background. 

I'm not sure if there is a good solution here besides using best judgment when implementing and/or designing sections.

robbieaverill:
Sweet as, fair enough

PR #389, opened 2020-12-17
Reduce left margin on small viewports for code block tabs to prevent line wrapping

robbieaverill:
Part of https://github.com/chec/marketing-chec/issues/481

If we're not happy with this adjustment then we'll need to get our hands dirty with the border radius on the background of the code block tabs. It's currently "rounded-full" which doesn't really work when it splits into two lines. Would need to be a specific number that works instead e.g. 200px or something.

This change prevents the home page on the marketing site from splitting over two lines.

### DIFF
--- a/src/components/CodeBlockGroup.vue
+++ b/src/components/CodeBlockGroup.vue
@@ -88,7 +88,11 @@ export default {
   @apply relative;
 
   &__tabs {
-    @apply absolute m-6 rounded-full overflow-hidden inline-block shadow-sm bg-gray-700 p-1 z-10;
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
LGTM

ScopeyNZ (approved):

robbieaverill:
I used a specific media query instead of `@screen` because (A) our linter complained about it and (B) it's specific to this problem

ScopeyNZ:
Yeah our linter complaining is a bug with the linter.

PR #392, opened 2020-12-22
ThumbnailCards are now all the same height as each other when used in a grid

robbieaverill:
Part of https://github.com/chec/dashboard/issues/534

### DIFF
--- a/src/components/ChecThumbnailCard.vue
+++ b/src/components/ChecThumbnailCard.vue
@@ -82,7 +82,7 @@ export default {
 
 <style lang="scss">
 .thumbnail-card {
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
It feels weird to make this `flex-col`. I'd have thought you could do this with `flex items-stretch`. If it works though then 🤷 .

ScopeyNZ (approved):
I wanted to use the Storybook to validate it works but every example only has one card. Oh well. I trust you.

ScopeyNZ:
Tbh I thought the CSS was being applied to the container, but now I realise it's just the card. Ignore my rambling

PR #403, opened 2021-01-15
[CHEC-1191] Update modal designs to match new borders, colours, text size, buttons

robbieaverill:


src/components/ChecModal/ChecModalHeader.vue
```diff
@@ -51,5 +51,9 @@ export default {
 <style lang="scss">
 .chec-modal-header {
   @apply flex justify-between text-gray-500 font-lato mb-8;
```
kvisca, line 53: Header color should be `text-gray-500`

kvisca (changes requested):

src/components/ChecModal/ChecModalHeader.vue
robbieaverill: Good spot

ScopeyNZ (approved):

PR #412, opened 2021-02-05
Fix bug where computed state changes were not being reflected in navigation

robbieaverill:


ScopeyNZ (approved):

src/components/ChecNavigation.vue
```diff
@@ -155,15 +155,19 @@ export default {
...
+      currentActiveSection: null,
       listItemFocus: null,
     };
   },
   computed: {
+    compiledTree() {
+      return this.tree.map(this.parseSection);
+    },
+    activeSection() {
+      return this.currentActiveSection || this.compiledTree[0];
```
ScopeyNZ, line 169: This computed var is being assigned on line 265. You'll need to convert this to the `get`/`set` style of computed attribute or change line 265 to update `currentActiveSection` instead.

PR #417, opened 2021-02-17
[CHEC-1300] Feature/slideout panel

jaepass:
Implemented UI lib with a sliding panel component
There were some transition issues that Guy helped to fix and I've been playing around with the durations and delays and think they're looking quite ok. Let me know what you guys think.
There are some weird kinks I can't quite figure out how best to implement when they panels start stacking, would love some feedback here. There's probably a programmatic way to do this not using styling classes but I can't quite figure out yet


src/components/ChecSlideoutPanel.vue
```diff
@@ -0,0 +1,170 @@
+<template>
+  <transition
+    appear
+    name="panel"
+    :duration="500"
+  >
+    <div class="slideout-panel">
+      <Overlay
+        v-if="renderOverlay"
```
robbieaverill, line 9: I think there should always be an overlay. I assume from the designs that it would be white (maybe the colour should be configurable) with a lower opacity, then if you stack them it would just overlay the previous one and have a slightly indented left side.

src/components/ChecSlideoutPanel.vue
```diff
@@ -0,0 +1,170 @@
...
+    size: {
+      type: String,
+      default: 'half',
+    },
+    /**
+     * The title of the panel slideout in the header
+     */
+    title: {
+      type: String,
+      required: true,
```
robbieaverill, line 57: This doesn't feel like it should be a requirement to me

src/components/ChecSlideoutPanel.vue
```diff
@@ -0,0 +1,170 @@
...
+     */
+    overlay: {
+      type: String,
+      default: 'light',
+      validator(value) {
+        return ['light', 'dark'].includes(value);
+      },
+    },
+    /**
+     * Whether to close panel on overlay click. Default is true.
```
robbieaverill, line 84: I think it actually defaults to false if you don't provide a default

src/components/ChecSlideoutPanel.vue
```diff
@@ -0,0 +1,170 @@
...
+    if (panels.length) {
+      this.renderOverlay = false;
+      panels.forEach((panel) => {
+        // eslint-disable-next-line no-param-reassign
+        panel.depth += 1;
+      });
+    }
+    panels.push(this);
+    console.log('Added', panels);
+  },
```
robbieaverill, line 106: I'm not too familiar just yet with how you've implemented this, but when I imagined it I was thinking that the panels don't need to know about each other. The "depth" prop is fine though, because you want to tell the panel how indented it should be (e.g. when it overlays another panel it has a depth of 2, etc). Any more logic than a number prop being used to add a CSS class feels like overkill to me. What do you think?

src/components/ChecSlideoutPanel.vue
```diff
@@ -0,0 +1,170 @@
...
+      /**
+       * Emitted when the close icon is clicked
+       * @event close
+       * @type {$event}
+       */
+      this.$emit('close', $event);
+    },
+    onOverlayDismiss() {
+      if (this.closeOnOverlayClick) {
+        this.$emit('close');
```
robbieaverill, line 130: `this.emitClose()` is better, then you are re-using existing logic

src/components/ChecSlideoutPanel/Overlay.vue
```diff
@@ -0,0 +1,50 @@
...
+     */
+    overlay: {
+      type: String,
+      default: 'light',
+      validator(value) {
+        return ['light', 'dark'].includes(value);
+      },
+    },
+    /**
+     * Whether to close panel on overlay click. Default is true.
```
robbieaverill, line 23:
```suggestion
     * Whether to close panel on overlay click. Default is false.
```

Or add `default: true` (the linter will complain if you do this though - maybe it needs to be `overlayClick` accepts strings e.g. `close` or `none`

src/components/ChecSlideoutPanel/Overlay.vue
```diff
@@ -0,0 +1,50 @@
...
+  },
+};
+</script>
+
+<style lang="scss">
+.slideout-panel {
+  &__overlay {
+    @apply fixed top-0 left-0 w-full h-full flex z-50 overflow-scroll transition-opacity ease-in-out duration-300;
+    // Special / overlay
+    background-color: rgba(255, 255, 255, 0.95);
```
robbieaverill, line 42: You can use `bg-white` and a background opacity class for this. Define a new opacity option in Tailwind config for 95% if you need to. https://v1.tailwindcss.com/docs/background-opacity#app

src/components/ChecSlideoutPanel/Overlay.vue
```diff
@@ -0,0 +1,50 @@
...
+
+    &--dark {
+      // Special / overlay
+      background-color: rgba(65, 85, 108, 0.9);
```
robbieaverill, line 46: Same here, let's use Tailwind classes and create new config for them if necessary

src/components/ChecSlideoutPanel/Panel.vue
```diff
@@ -0,0 +1,170 @@
...
+    ChecButton,
+    ChecIcon,
+  },
+  props: {
+    /**
+     * The size of the panel width. One of "half", "third", "threeQuarters", "full".
+     */
+    size: {
+      type: String,
+      default: 'half',
```
robbieaverill, line 57: I asked about using Tailwind widths in the other component - if you change it there then please do it here too. If not then ignore this.

src/components/ChecSlideoutPanel/Panel.vue
```diff
@@ -0,0 +1,170 @@
...
+    },
+    depth: Number,
+  },
+  computed: {
+    classNames() {
+      const { size } = this;
+      const { depth } = this;
+      return {
+        [`slideout-panel__element--size-${size}`]: size !== '',
+        [`slideout-panel__element--depth-${depth}`]: depth !== '',
```
robbieaverill, line 88: 👌

src/stories/components/ChecSlideoutPanel.stories.mdx
```diff
@@ -0,0 +1,338 @@
...
+                size="half"
+                overlay="light"
+                titleTag="h2"
+                width="2xl"
+                @close="openThird = false"
+              >
+                Hello, I'm the third panel!
+                <template v-slot:toolbar>
+                  <ChecButton @click="close" color="primary" text-only class="p-0">
+                    Cancel
```
robbieaverill, line 286: I don't like the way this is structured. I think what we want here is siblings rather than nesting:

```
<ChecSlideoutPanel ... />
<ChecSlideoutPanel ... />
<ChecSlideoutPanel ... />
```

z-index will mean they can naturally stack on top of each other, close when you click the overlay or close button, then reveal the next one in the list

src/components/ChecSlideoutPanel.vue
robbieaverill: Thinking more about this and chatting about it, it does actually make sense to do it this way. Some comments might help to clarify exactly why things are happening e.g. why we use an external variable over data property state.

src/stories/components/ChecSlideoutPanel.stories.mdx
jaepass: Will resolve this comment after discussing that nesting makes sense

jaepass:
All comments have been addressed aside from comments on the external variable panels and the nesting of the comments! One thing I couldn't quite address is adding widths to the stacking panels according to its depth as the panel widths are dynamic. Is there a way around this? Was trying to write a css function to get the width from a selector name. We could say the width should be 75% anytime its in a stacking group?

robbieaverill:
Yeah you can use negative left margins to achieve it probably

robbieaverill:
```
11:09:22 PM: ERR! => Failed to build the preview
11:09:22 PM: ERR! ./src/components/ChecSlideoutPanel.vue
11:09:22 PM: ERR! Module Error (from ./node_modules/eslint-loader/index.js):
11:09:22 PM: ERR! error: Unexpected console statement (no-console) at src/components/ChecSlideoutPanel.vue:105:9:
11:09:22 PM: ERR!   103 |         // eslint-disable-next-line no-param-reassign
11:09:22 PM: ERR!   104 |         panel.depth += 1;
11:09:22 PM: ERR! > 105 |         console.log('Added', panel);
11:09:22 PM: ERR!       |         ^
11:09:22 PM: ERR!   106 |       });
11:09:22 PM: ERR!   107 |     }
11:09:22 PM: ERR!   108 |     // Push a new panel component to the panels array
11:09:22 PM: ERR!
11:09:22 PM: ERR!
11:09:22 PM: ERR! 1 error found.
```

jaepass:
ah sorry

robbieaverill:
When no title is provided the "X" close button should still be right aligned:
image

src/components/ChecSlideoutPanel.vue
```diff
@@ -0,0 +1,172 @@
...
+      /**
+       * Emitted when the close icon is clicked
+       * @event close
+       * @type {$event}
+       */
+      this.$emit('close', $event);
+    },
+    onOverlayDismiss() {
+      if (this.closeOnOverlayClick) {
+        this.$emitClose('close');
```
robbieaverill, line 133:
```suggestion
        this.emitClose('close');
```

src/components/ChecSlideoutPanel/Overlay.vue
```diff
@@ -0,0 +1,60 @@
...
+      type: String,
+      default: 'light',
+      validator(value) {
+        return ['light', 'dark'].includes(value);
+      },
+    },
+    /**
+     * Whether to close panel on overlay click. Default is true.
+     */
+    closeOnOverlayClick: Boolean,
```
robbieaverill, line 25: The parent component has this prop already. I think this component can emit the event regardless, and then the parent component can check this prop and decide whether to do anything with the event.

src/components/ChecSlideoutPanel/Panel.vue
```diff
@@ -0,0 +1,173 @@
...
+      <div class="slideout-panel__header">
+        <div class="slideout-panel__header-inner">
+          <component :is="titleTag" v-if="title" class="slideout-panel__title">
+            {{ title }}
+          </component>
+          <ChecButton
+            class="slide-panel__close"
+            color="secondary"
+            variant="small"
+            @click="emitClose"
```
robbieaverill, line 19: I'd like to be able to press the Escape key and have it close the panel, could you add that too please?

src/components/ChecSlideoutPanel/Panel.vue
```diff
@@ -0,0 +1,173 @@
+<template>
+  <transition
```
robbieaverill, line 2: Looking at https://deploy-preview-417--chec-ui.netlify.app/?path=/story/components-slideout-panel--with-stacking-panels - when I open the first panel, it transitions the width (?) then looks like it's transitioning the left margin as well? Feels slightly odd, but I can definitely live with it.

ScopeyNZ:
I found a bug where opening and closing the second panel over and over again caused the first panel to constantly move more to the left. Something isn't resetting the position of the initial panel.

src/stories/components/ChecSlideoutPanel.stories.mdx
```diff
@@ -0,0 +1,377 @@
...
+              </ChecButton>
+            </div>
+          </template>
+        </ChecSlideoutPanel>
+        <ChecButton
+          @click="open = !open"
+          color="primary"
+          class="m-20"
+        >
+          Open the first panel
```
robbieaverill, line 366: After I click this, the button is still focussed. We should switch the focus when panels are opened away from whatever opened it. I can click this button then press enter repeatedly to toggle the panel's visibility. I think what should happen is that I press enter once and it opens up, then pressing enter again should trigger the "Save" button in the panel (assuming that the panel contains a form, and that the "Save" button is a submit button). Anyway - worth considering this.

src/components/ChecSlideoutPanel/Overlay.vue
```diff
@@ -0,0 +1,60 @@
...
+  &__overlay {
+    @apply
+      fixed
+      top-0
+      left-0
+      w-full
+      h-full
+      flex
+      z-50
+      overflow-scroll
```
robbieaverill, line 48: This is a bit weird IMO. I think when a panel is open that whatever was behind is should be static. You can disable scrolling behind modals by putting `overflow: hidden` on the body when a panel/modal/etc is open. Maybe check to see whether we already do something for this in the modal, and if so we might be able to re-use that. If we don't then we probably should and in this case we can probably tackle this as a separate issue and kill two birds with one stone.

robbieaverill (commented):
Nice work @jaepass 🎉  I've left a few more specific comments after doing a bit of testing

src/components/ChecSlideoutPanel/Panel.vue
jaepass: > Looking at https://deploy-preview-417--chec-ui.netlify.app/?path=/story/components-slideout-panel--with-stacking-panels - when I open the first panel, it transitions the width (?) then looks like it's transitioning the left margin as well? Feels slightly odd, but I can definitely live with it.
> 
> If this is caused by the depth calculation logic then I think it should only be modifying margins for nested panels, rather than a single panel on its own.

I'm still trying to find a better solution to dynamically add to the panel width 😓

jaepass:
> Looking at https://deploy-preview-417--chec-ui.netlify.app/?path=/story/components-slideout-panel--with-stacking-panels - when I open the first panel, it transitions the width (?) then looks like it's transitioning the left margin as well? Feels slightly odd, but I can definitely live with it.
> 
> If this is caused by the depth calculation logic then I think it should only be modifying margins for nested panels, rather than a single panel on its own.

I'm still trying to find a better solution to dynamically add to the panel width 😓

ScopeyNZ:
I went around in circles on this for a bit. I'm still not convinced about having these absolute floating panels potentially deeply nested in the DOM, but I can't get transitions and Vue Portal to play nicely.

I've done a couple of tweaks to fix some oddities I noticed but it was pretty much good to go. Happy with this now 🙂

PR #418, opened 2021-02-20
[CHEC-1311] Add filter bar component

ScopeyNZ:
Also:

- Add an "input" style for buttons that look like an input
- Allow custom IDs (and labels) for the switch component
- Fix missing BG colour on "white" tags
- Allow reset text to be customised in tag groups

This also closes [CHEC-1000]

See the designs I've been following here: https://www.figma.com/file/QapWtySLg7GOXEdGcTFVDZ/Global-UI?node-id=9075%3A12122

### DIFF
--- a/src/lang/en.js
+++ b/src/lang/en.js
@@ -14,6 +14,7 @@ export default {
       search: 'Search',
       // Appended to required form field labels
       requiredInline: '(required)',
+      // Used after other statements, like in the case of filters: "Refunded: yes"
       yes: 'yes',
     },
     fileManager: {
--- a/src/lang/en.js
+++ b/src/lang/en.js
@@ -12,11 +12,19 @@ export default {
     },
     general: {
       search: 'Search',
-      requiredInline: '(required)', // Appended to required form field labels
+      // Appended to required form field labels
+      requiredInline: '(required)',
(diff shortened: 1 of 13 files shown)


src/lang/en.js
```diff
@@ -12,12 +12,20 @@ export default {
     },
     general: {
       search: 'Search',
-      requiredInline: '(required)', // Appended to required form field labels
+      // Appended to required form field labels
+      requiredInline: '(required)',
+      yes: 'yes',
```
robbieaverill, line 17: Is this lowercase on purpose? If so, may be useful for translators if you provide a comment saying that and maybe why

robbieaverill (approved):
LGTM

src/lang/en.js
ScopeyNZ: 🤔 
It appears after a filter. For example on orders "Refunded: yes". Do you think I should:
- Put a comment saying it's used after other text like that example
- Put the comment in, and also move it out of "general" into "filters"
- Make it capitalised anyway (so it's "Refunded: Yes"
- Capitalise it, but transform it to lower case

robbieaverill: #1

PR #419, opened 2021-02-25
[CHEC-1310] Tags field with dropdown

jaepass:
- Added tags field with dropdown display the options added and a remove functionality
- Tags can be added with key events like 'enter', 'comma', 'tab'
- Tags field displays as an input on first add then renders in dropdown with subsequent adds
- There was no dropdown down arrow in design so made it to toggle on outside click

Kapture 2021-02-24 at 22 46 30


robbieaverill:
This design is confusing, what's the best way for me to provide feedback on that?

In terms of the implementation here I'd suggest initially that you disable the browser's autocomplete (if you can, I know it's hard to do sometimes).

ScopeyNZ:
@kvisca how did you expect this to work? In my mind, the dropdown is actually useless in this case - it should just let you type in and add things like when "small" is added in the example. The dropdown is useful when you want autocompletion, but it shouldn't show existing options in the field, and probably shouldn't render the the "remove" button...

kvisca:
> @kvisca how did you expect this to work? In my mind, the dropdown is actually useless in this case - it should just let you type in and add things like when "small" is added in the example. The dropdown is useful when you want autocompletion, but it shouldn't show existing options in the field, and probably shouldn't render the the "remove" button...

@ScopeyNZ the dropdown was added in the event we have long titled variants. @robbieaverill and I discussed this problem and came to the conclusion that if it's over say 9 characters, we shorten it and add an ellipsis in the pill. Then the dropdown can display the complete name and drop lines if need be. 

I can see how that is confusing though so maybe we remove the dropdown portion and just let them type to complete. We can remove the limit then from the pills and I'll think of a way to manage it with long variant titles.

jaepass:
Ok so I guess the consensus is to remove the dropdown for now in order to ship this. We can always revisit that later down the line. I do have a couple more questions @kvisca if you don't mind that I iron those out before refactoring this. Initially I had gone with referring to this UI http://www.vue-tags-input.com/#/ where it is just an input and added tags would be listed on the left. But after chatting about the designs with you, refactored to refer to your functional wireframe https://vue-multiselect.js.org/#sub-tagging

jaepass:
Kapture 2021-02-26 at 18 38 26

src/components/ChecTagsField.vue
```diff
@@ -0,0 +1,368 @@
...
+        // the input is not active
+        if (!this.tagsFieldFocused) {
+          this.activeInput = false;
+        }
+      }
+    };
+    document.addEventListener('keydown', onEscape);
+    this.$once('hook:destroyed', () => {
+      document.removeEventListener('keydown', onEscape);
+    });
```
ScopeyNZ, line 164: What's this? I haven't seen this sort of thing before. The `onOutsideClick` is the more standard way of doing this.

src/components/ChecTagsField.vue
```diff
@@ -0,0 +1,368 @@
...
+      if (!keyAddTag && (!this.addOnBlur || typeIsNotBlur)) {
+        return;
+      }
+
+      // Check to see if there is any callback function being passed
+      // before newTag is added or add 'newTag' entered
+      const tag = this.beforeAdding ? await this.beforeAdding(this.newTag) : this.newTag;
+
+      // If duplicates are allowed or if tag does not exist yet
+      if (tag && (this.allowDuplicates || this.options.indexOf(tag) === -1)) {
```
ScopeyNZ, line 208:
```suggestion
      if (tag && (this.allowDuplicates || this.options.includes(tag))) {
```

ScopeyNZ (changes requested):
Can you put a normal input next to this in the story? I think it's important to keep the empty (or one-line) inputs the same height - and I'm not sure they are right now.

Also, I think you should make this input "controlled" like the other inputs we have. This means the options should really be a `v-model`.

ScopeyNZ:
Let me know Jaeriah when you've addressed the feedback in my review 🙂

jaepass:
bah sorry @ScopeyNZ still trying to work through on hooking v-model corrrectly

ScopeyNZ:
Just going through this now. There's some super minor quirks I'm just trying to investigate - I'll ignore them if it's too finicky.

jaepass:
ok sounds good thanks Guy, let me know if you want me to look at anything

ScopeyNZ:
I've made some slight changes in a series of commits. Happy to go over them with you if you want to - let me know.

jaepass:
Thanks so much Guy, sorry you've had to come in to fix it up 😓 . I see what you've done with the field container and input, most times I never know which CSS properties to use to achieve a style correctly. Tested it, it's looking great!

ScopeyNZ:
Haha no sweat Jaeriah. It was getting pretty complicated so I wouldn't worry about it too much. That input auto-size thing is definitely a bit of a hack, and good ol' css-tricks helped me there.

ScopeyNZ (approved):

PR #420, opened 2021-02-25
A few updates/fixes

ScopeyNZ:
While working on new features in the dashboard I've found some bugs and wanted some new features:

- Add filter bar to exports
- Add a "snug" variant to tables
- Make z-index on panels lower so tooltips & popovers work
- Fix currency symbol in text fields without a label


ScopeyNZ:
🌲

PR #421, opened 2021-03-04
Integrating Localazy with UI library

kvisca:
You can test this locally by symlinking with the dashboard. I've sent out Localazy invites to everyone for this new project.


build/translations.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-disable-next-line import/no-extraneous-dependencies */
+const { exec } = require('child_process');
+
+const readKey = 'a8735366329350422768-8c5312346f2b785a5888e8488f383c20b02f12b909e57428e504c2c156c357e4';
+const writeKey = 'a8735366329350422768-20701159ffe67ba49888aa16aa6190439eb98683365f995d7788d95d3eadffae';
```
robbieaverill, line 5: Should these be in an environment file?

kvisca: Oops, yep your right.

ScopeyNZ:
You'll need to add a Netlify config file (or maybe an nvm file works) to ensure that Netlify can build it. Currently it's failing trying to use node 10 when the Localazy lib supports 11 at a minimum.

src/lang/meta.js
```diff
@@ -0,0 +1,35 @@
+const localazyMetadata = {
+  projectUrl: "https://localazy.com/p/chec-ui-library",
+  baseLocale: "en",
+  languages: [
```
ScopeyNZ, line 4: Does this error if you're missing the ja.js and fr.js files locally? In the dashboard this file is hand-written to only contain English as a language, so people don't have to run Localazy to have their dashboard work.

kvisca: @ScopeyNZ good catch, I overwrote what's supposed to be there while testing.  Restored.

PR #422, opened 2021-03-08
Partial revert of 8e490c59aef3569fa6649cca0d68121e17094858, adding some padding back

robbieaverill:
The padding was actually in the designs, and is important for the active state of menu items. I have added it back but at a size that looks even with the x padding.


kvisca (approved):
LGTM

PR #425, opened 2021-03-16
[CHEC-1357] Add functionality to the image manager component

ScopeyNZ:
There's two key updates here:

- You can now specify a list of existing images which powers the functionality of a "choose existing" button. 
- You can provide additional "actions" that can be performed on each image. Providing these will change the existing delete button into a more options menu

There's one more part to the image management UI in the designs - relating to providing fields for updating attributes on an individual image. I've been debating with myself whether this is a good idea to add to the UI library though. It's true that we can ecapsulate the functionality of showing the modal and displaying the image, but we'd need to provide a slot for the fields.

I'll discuss a little more with the team and follow up with another PR

### DIFF
--- a/src/assets/mixins.scss
+++ b/src/assets/mixins.scss
@@ -3,7 +3,7 @@
   3:4 -> aspect-ratio(3, 4),
   16:9 -> aspect-ratio(16, 9)
 */
-@mixin aspect-ratio($width, $height) {
+@mixin aspect-ratio($width, $height, $affectChildren: true) {
   @apply relative;
 
   &::before {
@@ -13,9 +13,11 @@
     padding-top: ($width / $height) * 100%;
   }
 
-  > img,
-  div {
-    @apply absolute w-full h-full inset-0 object-cover;
+  @if $affectChildren {
+    > img,
+    div {
+      @apply absolute w-full h-full inset-0 object-cover;
+    }
   }
 }
 
--- a/src/stories/components/ChecAvatar.stories.mdx
+++ b/src/stories/components/ChecAvatar.stories.mdx
@@ -1,8 +1,6 @@
-import { Meta, Props, Story, Preview } from '@storybook/addon-docs/blocks';
-import { action } from '@storybook/addon-actions';
+import { Meta, Preview, Props, Story } from '@storybook/addon-docs/blocks';
 import { select, text } from '@storybook/addon-knobs';
 import ChecAvatar from '../../components/ChecAvatar.vue';
-import TextField from '../../components/TextField.vue';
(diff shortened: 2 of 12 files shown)


src/assets/mixins.scss
```diff
@@ -13,9 +13,11 @@
...
   }
 
-  > img,
-  div {
-    @apply absolute w-full h-full inset-0 object-cover;
+  @if $affectChildren {
+    > img,
+    div {
+      @apply absolute w-full h-full inset-0 object-cover;
+    }
```
ScopeyNZ, line 20: In my use-case, I had a direct descendant that was absolutely positioned, which this was affecting. I don't think this definition should exist here, but I didn't want to adjust existing uses of this mixin.

ScopeyNZ:
I'll fix the linting error in a moment. I'm just going to turn it off as I think the order of the definitions in the CSS makes more sense than alphabetical.

src/stories/components/ChecAvatar.stories.mdx
```diff
@@ -70,7 +68,7 @@ import TextField from '../../components/TextField.vue';
       },
       props: {
         image: {
-          default: text('Image URL', 'http://lorempixel.com/400/400/people/'),
+          default: text('Image URL', 'https://www.fillmurray.com/400/400'),
```
robbieaverill, line 71: I can't see a license on this, we should use something that provides for open source repackaging

robbieaverill (approved):
LGTM, just a question about licensing on the placeholder image provider

src/stories/components/ChecAvatar.stories.mdx
ScopeyNZ: Good point. However we do use fillmurray already in the UI lib before this PR. `lorempixel` has been down for a while. Do you want me to go through and find something suitable and replace it everywhere it's in use?

robbieaverill: At some stage yeah, that'd be good

ScopeyNZ: Do you mind if I follow up on that? It looks like FillMurray uses creative commons images, and I can't find any that are different. We're probably fine using these as they are under CC now - provided they're used for layout only.

ScopeyNZ: I will merge and follow up later. I'd like to unblock dashboard work that uses these features.

robbieaverill: Yeah sorry, that's what I meant by later. All good!

PR #427, opened 2021-03-19
Fix UX issue with unclear active text search on filter bar

ScopeyNZ:
While @jaepass was testing an implementation of the filter bar on the dashboard, she noted some UX quirks around filtering. This PR adds tracking for the "active" text search in the filter bar, and controls to clear that search.

### DIFF
--- a/src/components/ChecFilterBar.vue
+++ b/src/components/ChecFilterBar.vue
@@ -1,12 +1,13 @@
 <template>
   <div class="filter-bar">
     <Search
+      ref="search"
       :filters="searchableFilters"
       :search="search"
       :disable-text-search="disableTextSearch"
+      :text-search-label="textSearchLabel"
       @add-filter="(filter) => addFilter(filter)"
       @search="setSearch"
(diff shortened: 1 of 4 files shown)


src/components/ChecFilterBar.vue
```diff
@@ -164,12 +179,24 @@ export default {
...
+      this.$emit('clear-search');
     },
     doSearch() {
       /**
        * Indicates that the user has requested a search should be performed with the current search term
        */
       this.$emit('search');
+
+      // Blur the search field to hide the popover
+      this.$refs.search.$el.querySelector('input').blur();
```
robbieaverill, line 199: Are we sure that this ref will exist? If not, do you think it's worth adding an if for type safety around it?

robbieaverill (approved):

src/components/ChecFilterBar.vue
ScopeyNZ: Yeah it'll always exist.

PR #429, opened 2021-03-21
[CHEC-997]Add a line chart component

ScopeyNZ:


ScopeyNZ:
I'd like to make sure this can be properly integrated before merge

ScopeyNZ:
Uhhh - I messed something up here.

ScopeyNZ:
This should be ready for review.

package.json
```diff
@@ -47,6 +48,7 @@
     "tiptap-extensions": "^1.33.1",
     "v-tooltip": "akryum/v-tooltip",
     "vue": "^2.6.10",
+    "vue-chartjs": "^3.5.1",
```
robbieaverill, line 51: Do we need to list chart.js as well as this library?

src/components/ChecChart/LineWrapper.vue
```diff
@@ -0,0 +1,240 @@
+<script>
+import { Line } from 'vue-chartjs';
+
+export default {
+  name: 'ChecLineChart',
```
robbieaverill, line 5: LineWrapper

src/components/ChecChart/LineWrapper.vue
```diff
@@ -0,0 +1,240 @@
...
+
+      let upperBound = 0;
+      let lowerBound = 0;
+
+      // TODO support more than just 2 lines
+      const setLowerBound = this.minimums[1];
+
+      if (secondSet) {
+        datasets.push({
+          label: 'Data Two',
```
robbieaverill, line 141: Does this label need to be here?

src/components/ChecLineChart.vue
```diff
@@ -0,0 +1,212 @@
...
+      this.tooltipOverlayPosition = {
+        left: `calc(${x} - ${padding} - ${bounds.x}px)`,
+        top: `calc(${y} - ${padding} - ${bounds.y}px)`,
+      };
+
+      // Figure out what index is being hovered over
+      const { index } = tooltipModel.dataPoints[0];
+      const title = tooltipModel.title[0];
+
+      this.tooltipContent = `<div class="chec-chart__tooltip"><div class="chec-chart__tooltip-title">${title}</div>${
```
robbieaverill, line 141: This doesn't look ideal, I assume there wasn't an easy way to do this with templates etc?

robbieaverill (approved):

src/components/ChecLineChart.vue
ScopeyNZ: Hmmn. I had a similar situation with the variants work that I found an alternative solution for. I'll try something similar here.

package.json
ScopeyNZ: https://github.com/chec/ui-library/pull/429/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R33

src/components/ChecLineChart.vue
ScopeyNZ: It looks like it's possible - but given the way I'm doing to tooltips right now (positioned on an invisible overlay) it might look just as hacky as what I've done already.

I'd prefer to just run with this for now - and deal with issues if they come up.

src/components/ChecChart/LineWrapper.vue
ScopeyNZ: Nope. Removed.

src/components/ChecChart/LineWrapper.vue
ScopeyNZ: Nice spotting. Fixed.

package.json
robbieaverill: Is this a link back to this comment? I'm confused

robbieaverill: I see you've highlighted a line - my question is whether you need to specify that line, since you're specifying vue-chartjs which presumably lists chart.js as a dependency already

ScopeyNZ: Oh sorry, I thought you were asking if I needed to add chart.js. Chart.js is listed as a peer dependency in `vue-chartjs` so yeah, I need `chart.js` in the package.json too.

PR #433, opened 2021-03-24
Adding toggle for opening code block icon links in a new tab

kvisca:
- ~~Adding toggle for opening code block icon links in a new tab~~
- Dynamically changing tag type depending on the link given. If no link is given, the tag defaults to span.

### DIFF
--- a/src/components/CodeBlockGroupIcon.vue
+++ b/src/components/CodeBlockGroupIcon.vue
@@ -1,8 +1,7 @@
 <template>
-  <a
-    :href="link"
-    target="_blank"
-    rel="noopener nofollow"
+  <component
+    :is="tag"
+    v-bind="tagProps"
   >
     <ReactLogo
       v-if="icon === 'react'"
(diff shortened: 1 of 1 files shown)


src/components/CodeBlockGroupIcon.vue
```diff
@@ -1,8 +1,8 @@
 <template>
   <a
     :href="link"
-    target="_blank"
-    rel="noopener nofollow"
+    :target="linkNewTab ? '_blank' : '_self'"
+    :rel="linkNewTab ? 'noopener nofollow' : ''"
```
robbieaverill, line 5: We need to use router-link for internal navigation. You can use a pattern match to see if the link is relative and either render an `<a>` or ` <router-link>` either way.

robbieaverill: For example: if link starts with `/` then use `<router-link :to="link">` otherwise use `<a :href="link" target="_blank" rel="noopener nofollow">`

kvisca: Updated, thanks for the suggestion.

robbieaverill (approved):
Looks good! Can you remove the comentes out code? Then good to merge

PR #434, opened 2021-03-24
Remove z-index on header and footer in panel

jaepass:
Sorry, need to revert the z-index as stacking panels were layered under


kvisca (approved):
LGTM

ScopeyNZ:
Just a note on this. `z-index` should ideally be completely avoided - unless it's really necessary. It's better to just ensure that thing appear later in the DOM to affect its "z" value.

PR #438, opened 2021-03-26
Fix bug where long data pill labels would break over multiple lines when in a group

robbieaverill:
Fixes https://github.com/chec/dashboard/issues/722 and fixes https://github.com/chec/dashboard/issues/739

### DIFF
--- a/src/components/ChecDataPillGroup.vue
+++ b/src/components/ChecDataPillGroup.vue
@@ -14,6 +14,10 @@ export default {
 .data-pill-group {
   @apply flex;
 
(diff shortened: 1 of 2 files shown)


ScopeyNZ:
Do you want them to break or not break? The PR makes it break anywhere, even in the middle of words.

robbieaverill:
Yeah I did this wrong. I think we need inline-block on the pills when in the group, will update.

ScopeyNZ (approved):
LGTM. I pushed an update to the storybook that I used to test it.

PR #445, opened 2021-03-30
Add tailwind plugin for hazard borders and add a story showing it off

ScopeyNZ:
I've built this in a way that will allow us to have some variants in the future if we want. The one right now uses a combination of grays so it's called "gray".

### DIFF
--- a/src/stories/styles/misc.stories.mdx
+++ b/src/stories/styles/misc.stories.mdx
@@ -1,0 +1,22 @@
+import { Meta, Story, Preview } from '@storybook/addon-docs/blocks';
+
+<Meta title="Styles/Misc" />
+
+# Miscellaneous additions
+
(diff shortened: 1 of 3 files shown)


kvisca (approved):
LGTM. I sort of agree that we should probably note that it's for horizontal use. Maybe we add horizontal to the name?

jaepass (approved):

jaepass:
Looks great to me!

ScopeyNZ:
> LGTM. I sort of agree that we should probably note that it's for horizontal use. Maybe we add horizontal to the name?

Yeah I was thinking about making it `border-x-hazard-gray`, but you can't specify a border image for a specific side.

Maybe we can just re-address the hazard styling when we want to put it somewhere other than the left and right?

PR #457, opened 2021-04-30
Add enabled pseudo to prevent hover style on disabled links

### LINKED ISSUE #369: Disabled "link" buttons still have a hover state background colour change
Shouldn't change colour on hover when a button of "link" type is disabled.

cc @blazepollard

jaepass:
Closes #369

### DIFF
--- a/src/components/ChecButton.vue
+++ b/src/components/ChecButton.vue
@@ -323,14 +323,14 @@ export default {
         @apply bg-#{map.get($config, 'default')} #{map.get($config, 'text')};
 
         &:not(.button--disabled):hover,
-        &:not(:disabled):hover {
+        &:not(:disabled):hover:enabled {
           // Override any gradient background
           background-image: none;
           @apply bg-#{map.get($config, 'hover')} #{map.get($config, 'text')};
         }
 
         &:not(.button--disabled):active,
-        &:not(:disabled):active {
+        &:not(:disabled):active:enabled {
           @apply bg-#{map.get($config, 'active')};
         }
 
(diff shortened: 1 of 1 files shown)


src/components/ChecButton.vue
```diff
@@ -323,14 +323,14 @@ export default {
         @apply bg-#{map.get($config, 'default')} #{map.get($config, 'text')};
 
         &:not(.button--disabled):hover,
-        &:not(:disabled):hover {
+        &:not(:disabled):hover:enabled {
```
robbieaverill, line 326: This seems kind of strange to me, does the `:not(:disabled)` part not work?

jaepass: Oddly no, not sure why it disregards the links as span tags but works for buttons

ScopeyNZ: Interesting, it's related to the different tag types I guess. Maybe the `:disabled` pseudo-selector doesn't make sense for anchor tags.

robbieaverill: Yeah anchors don't have that selector, good point. Maybe we should use classes instead?

ScopeyNZ (approved):
Seems to work okay to me. :shipit:

src/components/ChecButton.vue
ScopeyNZ: 🤷 I was just about to merge. It does look a bit weird having both, but CSS doesn't seem to have any issue and it works.

robbieaverill: That's fine

PR #462, opened 2021-04-30
Add defensive check for this.value existing before checking its length

robbieaverill:
Fixes https://github.com/chec/dashboard/issues/844


jaepass (approved):

ScopeyNZ (commented):
I think this has never been there in the past because the only valid types for the prop are string or array, but it's okay to be extra-defensive here I guess.

PR #465, opened 2021-06-15
Add a click event to the options menu button

ScopeyNZ:
I'm lazily loading the contents of the options menu, but it's not possible right now as I can't tell when the menu is being opened.


kvisca (approved):
LGTM

PR #472, opened 2021-07-05
Add missing disabled state to ChecSwitch component

robbieaverill:


ScopeyNZ (changes requested):
The switch goes green if it's disabled while _not_ toggled. I think that shouldn't happen.

robbieaverill:
Good catch, updated.

jaepass (approved):

ScopeyNZ (approved):

Issue #474, opened 2021-07-05
Loading z-index is higher than slideout panel


robbieaverill:
![image](https://user-images.githubusercontent.com/5170590/124524054-541a7480-ddae-11eb-93ce-3b192c9c7585.png)

Looks weird. The panel should be on top.

kvisca:
@robbieaverill can we close this based on the interaction on #475?

robbieaverill:
Nah it still needs to be resolved. This screenshot is an example of standard component use from the UI library and it shouldn't overlap like this. I closed #475 because there was a difference of opinion on how to fix this, but haven't spent any time on investigating whether we can remove the z-index from the Loading component yet. I think that's the way forward though - let the implementor decide what z-indexes will work for each use case (if they need to at all)

ScopeyNZ:
Ultimately we should probably have a documented z-index strategy. Different levels should have different purposes. For example:

- `z-0` (none): Normal content
- `z-10`: Sectional loading indicators
- `z-20`: Panels and modals
- `z-30`: Toasts and banner alerts
- `z-40`: Full page loading indicators

We could even rename the `z-` options in tailwind to be clear on the use of each z index, so you'd do something like `@apply z-modal` for modals.

ScopeyNZ:
As Robbie mentioned, sectional loading indicators shouldn't actually have z-index, but it might be a bit of a job to do that, as then the loading indicator MUST appear after the content it overlays in the DOM.

PR #475, opened 2021-07-05
Increase z-index on slideout panel so it appears on top of loading indicators (etc)

robbieaverill:
Fixes https://github.com/chec/ui-library/issues/474


ScopeyNZ (commented):
This seems to me like it should be optional functionality, as I can imagine scenarios where you want the loading indicator to appear above panels. Can you make it optional?

robbieaverill:
I don't think that the loading indicator should be specifying a z-index at all. Are you happy for me to remove it from that component?

ScopeyNZ:
Hmmmn. I like this in theory, but I'm not sure if it's going to be easy to do in practice. We might rely on the z-index too much already. If you want to give it a quick test then I'm happy for you to do that.

PR #479, opened 2021-07-09
Fix bug with icons inside buttons in alerts being given incorrect text colour

robbieaverill:
Ideally this uses BEM classes instead of `svg` targetting specifically, but the ChecIcon component has a custom render method and I didn't want to spend too much time trying to get custom classes to be passed through, so this is a quick fix.


jaepass (approved):

src/stories/components/ChecAlert.stories.mdx
```diff
@@ -172,3 +173,36 @@ import ChecIcon from '../../components/ChecIcon.vue';
...
+            variant="info"
+            inline
+          >
+            <ChecIcon icon="info-square" />
+            Hello world, I am some text.
+            <ChecButton
+              color="primary"
+              variant="round"
+              icon="info-square"
+              class="float-right"
```
robbieaverill, line 200:
```suggestion
```

PR #498, opened 2021-08-06
Add focus event for keyboarding

### LINKED ISSUE #484: TagsField input is not tabbable
Reproducible on the storybook. The remove buttons for each tag are correctly tabbable, but the actual input field to add new options is NOT tabbable. Annoying!

### LINKED ISSUE #500: Tabbing out of a tag field retains focus on the field
Tabbing out of the field will retain a focus state on it, even when I type into a different field

https://user-images.githubusercontent.com/3260989/128578849-7cc3f1eb-bc2d-47f8-87f3-200188b75896.mov

jaepass:
Fixes #484 
Fixes #500 

Silly me, originally went for this unconventional fix and took me longer than it should have. Fixed now

### DIFF
--- a/src/components/ChecTagsField.vue
+++ b/src/components/ChecTagsField.vue
@@ -238,11 +238,12 @@ export default {
       // input value is automatically added as a new tag.
       this.handleAddNewTag(e);
     },
+    /**
+     * Handler to set tags field and input to true when the dismiss button is focused
+     */
     handleDismissFocused() {
       this.tagsFieldFocused = true;
-      if (this.tagsFieldFocused) {
-        this.isInputVisible = true;
-      }
+      this.isInputVisible = true;
     },
     /**
      * Remove tag from tags list
--- a/src/components/ChecTagsField.vue
+++ b/src/components/ChecTagsField.vue
@@ -16,6 +16,7 @@
         <ChecTag
           :active="activeTag"
           :disabled="disabled"
+          @focused="handleDismissFocused"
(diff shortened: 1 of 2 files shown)


src/components/ChecTagsField.vue
```diff
@@ -237,6 +238,12 @@ export default {
       // input value is automatically added as a new tag.
       this.handleAddNewTag(e);
     },
+    handleDismissFocused() {
+      this.tagsFieldFocused = true;
+      if (this.tagsFieldFocused) {
```
ScopeyNZ, line 243: Won't this never be `false`?

jaepass: Yes true, i'll just take out the conditional

ScopeyNZ:
I've done some testing, and this component has a few issues 😅 . I've compared them to https://chec-ui.netlify.app though, so I'll only mention things that aren't previously issues:

You can't delete tags in an unfocused field (if showing the input will change the size of the field):

https://user-images.githubusercontent.com/3260989/128578772-205ed120-0054-485c-8599-3f4ce2121118.mov

I've also spotted #499, but interestingly the behaviour is different in this PR as the tag you click on isn't actually deleted:

https://user-images.githubusercontent.com/3260989/128579095-a149b077-2076-4597-85a4-06ded75079c7.mov

jaepass:
Thanks Guy!

ScopeyNZ (approved):

ScopeyNZ:
Nice :)

PR #514, opened 2021-09-03
Remove flex to fix overflow issue in inner block

jaepass:
Fixes https://github.com/chec/dashboard/issues/1172
Screen Shot 2021-09-03 at 4 07 12 PM

### DIFF
--- a/src/components/ChecCard/InnerBlock.vue
+++ b/src/components/ChecCard/InnerBlock.vue
@@ -75,7 +75,7 @@ export default {
 <style lang="scss">
 .card-inner-block {
   @apply
-    flex flex-row items-center justify-between w-full bg-gray-100 rounded p-4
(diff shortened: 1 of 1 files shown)


src/components/ChecCard/InnerBlock.vue
```diff
@@ -75,7 +75,7 @@ export default {
 <style lang="scss">
 .card-inner-block {
   @apply
-    flex flex-row items-center justify-between w-full bg-gray-100 rounded p-4
+    flex-row items-center justify-between w-full bg-gray-100 rounded p-4
```
ScopeyNZ, line 78:
```suggestion
    w-full bg-gray-100 rounded p-4
```

Theoretically all those don't matter if it's not `display: flex`.

PR #520, opened 2021-09-23
Refactor to rollup & update storybook

ScopeyNZ:
The only bundle offered by the UI lib after this will be an ESM bundle. This should be fine as both the dashboard and marketing site environments should prefer ESM bundles anyway. I've got a dashboard update PR ready to go, and I'll prep a marketing-site one now.

I'll release this as version 0.1.0 (we won't get to 0.0.100 lol). I've already done 9 "alpha"s as part of testing this update.


ScopeyNZ:
I've just spotted that the datepickers are broken - I commented out flatpickr css but forgot to re-add. Shouldn't take a moment.

rollup.config.js
```diff
@@ -0,0 +1,64 @@
...
+// PostCSS plugins
+import tailwindcss from 'tailwindcss';
+import autoprefixer from 'autoprefixer';
+import assets from 'postcss-assets';
+
+module.exports = {
+  input: {
+    index: 'src/index.js',
+    ChecLineChart: 'src/components/ChecLineChart.vue',
+    ChecWysiwyg: 'src/components/ChecWysiwyg.vue',
```
robbieaverill, line 21: Why must these two components be listed individually? Could you add a comment to say why?

robbieaverill (approved):

rollup.config.js
ScopeyNZ: Yeah it helps with code splitting on the UI library, but I don't think it's an amazing solution. Have put a comment in, but I'll also raise an issue to improve this later.

PR #544, opened 2021-11-02
Edit incorrect required prop, add error state

jaepass:
Not sure what I was thinking with this last pr https://github.com/chec/ui-library/pull/543 on adding required styling. This PR is a follow up to fix that and add an error state
Screen Shot 2021-11-02 at 1 37 07 PM


kvisca (approved):
LGTM

PR #546, opened 2021-11-04
[CHEC-1834] Add new radio group component, add stories, delete single radio story

### LINKED ISSUE #535: Label prop not taking in value
Spotted while trying to integrate it, the `label` prop does not take in a value, always returns undefined
Screen Shot 2021-10-14 at 10 53 17 PM

To reproduce:
- Consume ChecRadio or ChecCheckbox somewhere 
- Use a random value on the `label` prop

jaepass:
1/2 of task, part 2 is refactoring any implementation that's using ChecRadio. I think it's just attributes in dashboard and the marketing contact form. Will do that once this is merged.

Screen Shot 2021-11-04 at 9 45 15 AM

Screen Shot 2021-11-04 at 9 45 24 AM
5 

Fixes #535


robbieaverill (approved):

ScopeyNZ (commented):
Nice. We probably didn't need to drop the radio story, the component can still be used on it's own, but oh well.

jaepass:
Oh shucks, I can add it back in a follow up!

ScopeyNZ:
Na don't worry about it.

PR #553, opened 2021-11-04
Fixing an issue where the radio options label would be rendered as undefined

kvisca:
ChecRadioGroups radio options would be rendered as undefined because of some old code that required the label to be passed as a slot.

### DIFF
--- a/src/components/ChecRadio.vue
+++ b/src/components/ChecRadio.vue
@@ -11,15 +11,7 @@
       @input="handleInput"
     >
     <span class="chec-radio__fill" />
-    <!--
-      @slot Custom label slot
-      @bind label string
-      @bind isChecked boolean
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
I still really don't understand why this change actually does anything. The value inside a slot is supposed to be a default.

Did you double check that there are no instances of radio in use that use the slot already?

Also, can you please delete the comment above that refers to the slot?

kvisca:
I believe it has something to do with how the slot is set up. It's causing a conflict with the label. Besides, now that we have the `ChecRadioGroup` we shouldn't be using it on its own again and it makes more sense to pass as a prop instead of a slot. 

Removed the comments as requested.

ScopeyNZ (approved):

PR #554, opened 2021-11-08
[CHEC-1785] Updating green CTA buttons as they fail WCAG compliance

kvisca:
- Updating the green button instances to make them WCAG compliant.
- Fixed a few errors I saw.


src/components/ChecDataPill.vue
```diff
@@ -66,6 +66,14 @@ $colors: 'green', 'orange', 'purple', 'red', 'blue';
     }
   }
 
+  &--green {
+    @apply bg-green-100 text-green-700;
+
+    &.data-pill--filled {
+      @apply bg-green-600 text-white;
```
ScopeyNZ, line 73: Is there a reason we didn't just change the 600 and 500 values to work, rather than this? It feels a bit weird that green is the only colour with a 700 value now.

kvisca: I did this in the event we still want to use that color in other places. I'll update to remove the original and shift it down.

kvisca: @ScopeyNZ updated.

ScopeyNZ (approved):

ScopeyNZ:
Yeah I think just forcing the change in green everywhere is best.

PR #572, opened 2021-11-25
Allow tooltips to have line breaks with i18n json

kvisca:
Allow tooltips to have line breaks with i18n json using `/n`

### DIFF
--- a/src/assets/tooltips.scss
+++ b/src/assets/tooltips.scss
@@ -7,7 +7,7 @@
   display: block !important;
 
   .tooltip-inner {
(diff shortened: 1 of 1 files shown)


robbieaverill (approved):

robbieaverill:
Before merging, I have a feeling that tooltips support slots - if that's the case we probably don't need this. Happy to merge anyway though as it seems pretty harmless.

kvisca:
> 

They do support slots, but it may be a bit weird since it's a tooltip prop on the ChecFormField. We could refactor ChecFormField to support slotted tooltips instead. This change should be pretty harmless though.

Issue #562, opened 2021-11-17
Apostrophes not rendering correctly in storybook


eufandem:
<img width="478" alt="Screen Shot 2021-11-16 at 4 55 18 PM" src="https://user-images.githubusercontent.com/29156467/142089724-a8728f29-6628-4eb6-9f91-0562eb0ec89e.png">

https://chec-ui.netlify.app/?path=/story/components-notifications--with-button

kvisca:
@eufandem is this still happening to you? It seems to be okay for me. 

![Screen Shot 2021-11-26 at 1 23 31 PM](https://user-images.githubusercontent.com/36721153/143619356-3c0166fd-4c61-40d1-8a2f-830382057f9d.png)

robbieaverill:
I think this was from typing into the slot to change the content in the story:
![image](https://user-images.githubusercontent.com/5170590/143619601-edd51a3a-a879-446d-9f4a-98952bbd58a4.png)

PR #575, opened 2021-11-26
Fix bug where apostrophes were HTML encoded in ChecAlert storybook example

robbieaverill:
Fixes https://github.com/chec/ui-library/issues/562


kvisca (approved):
LGTM

PR #577, opened 2021-11-26
Increase z-index of panel

### LINKED ISSUE #565: Loading indicator z-index is higher than slideout panel mask
image

I guess it should be lower than the slideout panel's mask z-index

jaepass:
With https://github.com/chec/dashboard/pull/1488 fix, the slideout panel would be under the nav header and sidebar. This PR fixes that.

Fixes #565 as well

### DIFF
--- a/src/components/ChecLoading.vue
+++ b/src/components/ChecLoading.vue
@@ -59,7 +59,7 @@ export default {
 
 <style lang="scss">
 .loading {
-  @apply absolute top-0 left-0 w-full h-full z-50 flex justify-center items-center flex-col;
+  @apply absolute top-0 left-0 w-full h-full flex justify-center items-center flex-col;
   border-radius: inherit;
 
   &__animation {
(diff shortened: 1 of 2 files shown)


robbieaverill (approved):

ScopeyNZ:
Do tooltips, dropdowns, and loading indicators still work in panels?

jaepass:
Alright so latest push fixes loading indicators to only have z-index if it has a background, which we mostly use for full page loads, otherwise it would not have a z-index. All tooltips, dropdowns work as expected.

Example of loading indicator without bg displaying as expected.
Screen Shot 2021-11-26 at 12 12 05 PM

ScopeyNZ (approved):

ScopeyNZ:
I'm always nervous about these arbitrary z-index changes. Should be pretty obvious if this negatively affects something else though.

jaepass:
Yeahup for sure, will keep an eye out TA

PR #578, opened 2021-11-26
Removing hover, focus and active from tags

### LINKED ISSUE #555: ChecTag: should not have a hover state unless it is interactive
The ChecTag component has a hover state where it switches from light to dark (with default props). It shouldn't do this because clicking on the tag does nothing by default.

kvisca:
Realized that this isn't actually a button so I removed the hover states outright. Issue confused me but I can't find an instance where these are clickable as a router link. 

Closes #555

### DIFF
--- a/src/components/ChecTag.vue
+++ b/src/components/ChecTag.vue
@@ -88,10 +88,6 @@ export default {
     top: 2px;
   }
 
-  &--active {
-    @extend %active;
-  }
-
   &--disabled {
     @apply opacity-40 bg-white text-gray-500 cursor-not-allowed;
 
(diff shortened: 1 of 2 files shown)


robbieaverill:
I think the blog post in the marketing site uses them as links. We could refactor this to use tags.

kvisca:
@robbieaverill yea that's weird, they're just wrapped in a link. I'm going to make a button-type tag and replace it. PR incoming.

kvisca:
Opened PR fixing in chec/marketing-chec#714

robbieaverill (approved):
LGTM, there's an "active" knob in one of the tag stories that could be removed as well if you want to delete it

robbieaverill:
cc @kvisca

kvisca:
@robbieaverill sorry about that, I'll update that right now.

robbieaverill (approved):

PR #579, opened 2021-11-26
Enforce a max option height for dropdowns

ScopeyNZ:
Put a reasonable max height for dropdowns that doesn't just rely on viewport size

### DIFF
--- a/src/components/ChecDropdown.vue
+++ b/src/components/ChecDropdown.vue
@@ -551,7 +551,9 @@ export default {
 
   &__options {
     @apply overflow-y-auto overflow-x-hidden;
-    max-height: 40vh;
+
+    // 45rem is ~15 rows high as each row is a tiny bit less than 3rem without borders, and the borders add a bit more
+    max-height: min(40vh, 45rem);
   }
 
   &__down-arrow {


src/components/ChecDropdown.vue
```diff
@@ -551,7 +551,9 @@ export default {
 
   &__options {
     @apply overflow-y-auto overflow-x-hidden;
-    max-height: 40vh;
+
+    // 45rem is ~15 rows high as each row is a tiny bit less than 3rem without borders, and the borders add a bit more
+    max-height: min(40vh, 45rem);
```
robbieaverill, line 556: Do we need 40vh here? 45rem seems reasonable

ScopeyNZ: I left it so that it doesn't change the existing behaviour. I'm not sure why it was there in the first place though

ScopeyNZ: On second thought, I think it's there (below 50vh) to ensure it's not higher than half the viewport, which would mean you'd need to scroll the page to see all the options.

robbieaverill (approved):

PR #581, opened 2021-11-30
Allow accordions to be toggled when clicked. Make emitted event consistent

ScopeyNZ:
This was annoying me in the FAQ on the docs site.

### DIFF
--- a/src/components/ChecAccordion.vue
+++ b/src/components/ChecAccordion.vue
@@ -4,26 +4,28 @@
     class="accordion"
     :class="{'accordion--active': isOpen}"
   >
-    <div class="accordion__heading">
+    <div class="accordion__heading" @click="handleHeadingClick">
       <div>
         <div class="accordion__title" v-html="title" />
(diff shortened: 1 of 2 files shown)


src/components/ChecAccordion.vue
```diff
@@ -4,26 +4,28 @@
     class="accordion"
     :class="{'accordion--active': isOpen}"
   >
-    <div class="accordion__heading">
+    <div class="accordion__heading" @click="handleHeadingClick">
```
robbieaverill, line 7: Do you need to give it a role attribute for keyboard accessibility?

ScopeyNZ: Nope. The keyboard should still highlight the actual switch or button, I don't think we want the header to be tab-navable.

ScopeyNZ: Alternatively I can make the header/accordion tabbable, and make the button not tabbable - but that's a bunch more work

PR #582, opened 2021-12-03
Update dropzone to new major version

ScopeyNZ:
The new major version fixes a bug where the bundle relies on window (they've dropped "autodiscover" which did this). 

They've also dropped IE support, but I think we're okay with this.


ScopeyNZ:
I've added this as we can't use the UI lib in the marketing site right now, as the outputted bundle relies on window.

robbieaverill (approved):

PR #583, opened 2021-12-03
Make usage of portals optional when using ChecPopover

ScopeyNZ:
Due to the [caveats](https://portal-vue.linusb.org/guide/caveats.html) of `portal-vue`, you might just want to use the "popper" functionality of a popover, without moving the DOM to a root node. This makes that possible.


ScopeyNZ:
I didn't mean to add that `ChecButton` change yet. Will back that out

robbieaverill (approved):
Merge after reverting the button stuff

PR #586, opened 2021-12-06
New designs for notifications & rename ChecNotification to ChecToast

dvnkshl:
image


src/components/ChecToast.vue
```diff
@@ -16,16 +25,18 @@
...
   components: {
     SvgCloseIcon,
+    ChecIcon,
   },
   props: {
     /**
-     * The style variant for the notification. One of "success", "error", "warning", "info". Default is "info"
-     */
+         * The style variant for the notification. One of "success", "error", "warning", "info". Default is "info"
+         */
```
ScopeyNZ, line 39: Your IDE is doing something strange here I'm guessing? These comments were correct before.

src/components/ChecToast.vue
```diff
@@ -34,17 +45,27 @@ export default {
...
+         * The time in milliseconds after which a notification will disappear, if `persist` is true
+         */
     hideTime: {
       type: Number,
       default: 5000, // milliseconds
     },
   },
+  data() {
+    return {
+      icon: {
```
ScopeyNZ, line 61: This is a constant and shouldn't be defined in `data`. Variables created here will be reactive, which is unnecessary for this.

ScopeyNZ (changes requested):
I'll have a look at the deploy preview in a moment.

src/components/ChecToast.vue
ScopeyNZ: How I'd recommend you change this is just create a `computed` that does something like:

```
const icons = {...};
return icons[this.variant];
```

src/components/ChecToast.vue
dvnkshl: yeah weird fixed - locally

ScopeyNZ:
The hover state on the button is broken: 

image

The button style is wrong, as there seems to be some font styling that's overwriting the button font, but I guess that this is a pre-existing issue.

dvnkshl:
@ScopeyNZ updated

src/components/ChecToast.vue
```diff
@@ -80,29 +101,36 @@ export default {
...
-      @apply mt-2;
+      @apply mt-3 -ml-1;
     }
+
   }
 
   &__close-icon {
-    @apply ml-4 p-2 bg-green-600 rounded text-white cursor-pointer;
+    @apply bg-green-600 rounded text-white cursor-pointer mr-2 mt-4 p-2;
+    transition: all 0.3s ease, outline 0ms !important;
```
ScopeyNZ, line 131: Why do we have an `!important` here?

src/components/ChecToast.vue
```diff
@@ -80,29 +101,36 @@ export default {
...
 
-    > svg {
+    >svg {
```
ScopeyNZ, line 133: This was better the way it was before...

src/components/ChecToast.vue
dvnkshl: Without !important

 

With !important

ScopeyNZ: Right, if you want to get rid of the outline, just make sure it's `outline: none`.

ScopeyNZ (approved):

src/stories/components/ChecNotification.stories.mdx
```diff
@@ -2,19 +2,19 @@ import { Meta, Props, Story, Preview } from '@storybook/addon-docs/blocks';
 import { action } from "@storybook/addon-actions";
 import { text, select } from '@storybook/addon-knobs';
 import ChecButton from '../../components/ChecButton.vue';
-import ChecNotification from '../../components/ChecNotification.vue';
+import ChecToast from '../../components/ChecToast.vue';
 
-<Meta title="Components/Notifications" component={ChecNotification} />
+<Meta title="Components/Toast " component={ChecToast} />
```
robbieaverill, line 7: Next time you're in here please rename the file as well

PR #589, opened 2021-12-07
Edit shadow prop to have shadow by default

jaepass:
Will update dashboard and marketing site after this merge and release

### DIFF
--- a/src/components/ChecCard.vue
+++ b/src/components/ChecCard.vue
@@ -79,9 +79,9 @@ export default {
      */
     hoverable: Boolean,
     /**
-     * Allows card to have shadow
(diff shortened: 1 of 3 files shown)


ScopeyNZ (approved):
:chefs-kiss:

jaepass:
Yeesh and sorry with my past version tags. I'll tag the next one as 0.2.11 since the last was technically 0.2.10?

robbieaverill:
The last tags were 0.2.0 and 0.2.1, we can go with 0.2.2 up as far as we need to (e.g. 0.2.123543534), but no big deal.

PR #590, opened 2021-12-09
Fix bug where multiline text in toast messages had incorrect vertical padding

robbieaverill:


ScopeyNZ (changes requested):
I assume this is unintentional:

image

robbieaverill:
The escaping or the button being on the right? Storybook escapes HTML by default

robbieaverill:
Updated

ScopeyNZ (approved):
Hawt

PR #591, opened 2021-12-17
Remove dist files, ignore them, add action to release package instead

ScopeyNZ:
With this, the workflow to publish the UI library becomes:

- Navigate to the actions in GitHub 
 
image

- Click the on the publish workflow, and then click the "run workflow" button

image

- Enter the new version

image

- Hit the run button


.github/workflows/publish.yml
```diff
@@ -0,0 +1,27 @@
+name: Publish
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Enter a valid semver version'
+        required: true
```
robbieaverill, line 8: What does this do exactly?

robbieaverill (approved):

.github/workflows/publish.yml
ScopeyNZ: Soz, was just updating the OP with images. Raising PRs from the command line is convenient when you don't want to add images :P

ScopeyNZ:
Actually I've forgotten to specify the NPM token here. One moment

.github/workflows/publish.yml
robbieaverill: Nice, looks good

PR #592, opened 2021-12-18
Add pie chart

robbieaverill:
image

Designs:
image

### DIFF
--- a/src/components/PieChart.vue
+++ b/src/components/PieChart.vue
@@ -1,0 +1,93 @@
+<template>
+  <div class="pie-chart">
+    <h3 class="pie-chart__title">
+      {{ title }}
+    </h3>
+    <div class="pie-chart__chart-container">
+      <PieWrapper
+        class="pie-chart__chart"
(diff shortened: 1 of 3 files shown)


src/components/PieChart.vue
```diff
@@ -0,0 +1,93 @@
...
+  },
+  props: {
+    title: String,
+    dataPoints: Array,
+  },
+  data() {
+    return {
+      colors: {
+        green: 'rgb(52, 202, 173)', // green-300
+        purple: 'rgb(123, 101, 205)', // purple-400
```
ScopeyNZ, line 35: Minor, but this could be a prop with a default value in case we ever want other pie graphs with more than 2 points.

ScopeyNZ (approved):
Nice

src/components/PieChart.vue
robbieaverill: Yeah I figure that if we add more points we would use the other colours e.g. red, orange, blue, etc. We can add them as needed.

